PR #1, opened 2022-11-23
Create project structure

dfed:
This PR creates the project structure, setting up the:
* Package.swift file and subdirectories
* CI with code coverage
* Basic issue template
* Contributing.md
* Podspec
* Gemfile


Package.swift
```diff
@@ -0,0 +1,27 @@
...
+
+import PackageDescription
+
+let package = Package(
+    name: "swift-async-queue",
+    platforms: [
+        .macOS(.v10_15),
+        .iOS(.v13),
+        .tvOS(.v13),
+        .watchOS(.v6),
```
dfed, line 12: We're only running CI on the latest supported versions right now. I don't want to support Xcode versions prior to 14.1.

Scripts/build.swift
```diff
@@ -0,0 +1,219 @@
...
+        throw TaskError.code(task.terminationStatus)
+    }
+}
+
+enum TaskError: Error {
+    case code(Int32)
+}
+
+enum Platform: String, CaseIterable, CustomStringConvertible {
+    case iOS_13
```
dfed, line 24: We aren't using these older versions in CI, but they can be useful for running tests on older simulator OS versions locally.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,68 @@
...
+        run: bundle install
+      - name: Select Xcode Version
+        run: sudo xcode-select --switch /Applications/Xcode_14.1.0.app/Contents/Developer
+      - name: Build and Test Framework
+        run: Scripts/build.swift ${{ matrix.platforms }}
+      - name: Prepare Coverage Reports
+        run: ./Scripts/prepare-coverage-reports.sh
+      - name: Upload Coverage Reports
+        if: success()
+        uses: codecov/codecov-action@v3
```
dfed, line 53: I'm surprised we're not yet getting any code coverage posts just yet, but we also have nearly zero code in the repo. I may put in a pin in this and hope it gets fixed by adding code or running this code on `main`. I'll fix this if either of the prior do not work.

dfed:
YOLO-merging this scaffolding PR. Future PRs with real content will receive reviews.

.github/workflows/ci.yml
dfed: It's working! https://github.com/dfed/swift-async-queue/pull/2#issuecomment-1324577222

PR #2, opened 2022-11-23
Create AsyncQueue

dfed:
This PR introduces the `AsyncQueue` class, which makes it possible to send ordered work to an asynchronous context from a synchronous or non-isolated context.

Any and all API feedback is welcome! Note that the name of the queue changes in #3, which builds upon this work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![License](https://img.shields.io/cocoapods/l/swift-async-queue.svg)](https://cocoapods.org/pods/swift-async-queue)
 [![Platform](https://img.shields.io/cocoapods/p/swift-async-queue.svg)](https://cocoapods.org/pods/swift-async-queue)
 
-A queue that enables ordered sending of events from synchronous to asynchronous code.
+A queue that enables sending FIFO-ordered tasks from synchronous to asynchronous contexts.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,31 @@
 # swift-async-queue
-A queue that enables ordered sending of events from synchronous to asynchronous code
+[![Swift Package Manager compatible](https://img.shields.io/badge/SPM-compatible-4BC51D.svg?style=flat)](https://github.com/apple/swift-package-manager)
+[![codecov](https://codecov.io/gh/dfed/swift-async-queue/branch/main/graph/badge.svg?token=nZBHcZZ63F)](https://codecov.io/gh/dfed/swift-async-queue)
+[![Version](https://img.shields.io/cocoapods/v/swift-async-queue.svg)](https://cocoapods.org/pods/swift-async-queue)
+[![License](https://img.shields.io/cocoapods/l/swift-async-queue.svg)](https://cocoapods.org/pods/swift-async-queue)
+[![Platform](https://img.shields.io/cocoapods/p/swift-async-queue.svg)](https://cocoapods.org/pods/swift-async-queue)
+
+A queue that enables ordered sending of events from synchronous to asynchronous code.
+
+## Usage
+
+### Basic Initialization
+
+```swift
+let asyncQueue = AsyncQueue()
+```
+
+### Sending events from a synchronous context
+
+```swift
+asyncQueue.async { /* awaitable context that executes after all other enqueued work is completed */ }
+```
+
+### Awaiting work from an asynchronous context
+
+```swift
+await asyncQueue.await { /* throw-able, return-able, awaitable context that executes after all other enqueued work is completed */ }
+```
 
 ## Requirements
 
--- a/Scripts/build.swift
+++ b/Scripts/build.swift
@@ -174,6 +174,9 @@ for rawPlatform in rawPlatforms {
     if platform.shouldTest {
         xcodeBuildArguments.append("test")
     }
+    xcodeBuildArguments.append("-test-iterations")
+    xcodeBuildArguments.append("100")
+    xcodeBuildArguments.append("-run-tests-until-failure")
 
     try execute(commandPath: "/usr/bin/xcodebuild", arguments: xcodeBuildArguments)
     isFirstRun = false
--- a/Scripts/build.swift
+++ b/Scripts/build.swift
@@ -159,7 +159,8 @@ for rawPlatform in rawPlatforms {
         "-scheme", "swift-async-queue",
         "-sdk", platform.sdk,
         "-derivedDataPath", platform.derivedDataPath,
-        "-PBXBuildsContinueAfterErrors=0"
+        "-PBXBuildsContinueAfterErrors=0",
+        "OTHER_SWIFT_FLAGS=-warnings-as-errors",
     ]
 
     if !platform.destination.isEmpty {
--- a/Sources/AsyncQueue/AsyncQueue.swift
+++ b/Sources/AsyncQueue/AsyncQueue.swift
@@ -20,4 +20,70 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-final class AsyncQueue {}
+/// A queue that enables ordered sending of events from synchronous to asynchronous code
+public final class AsyncQueue: Sendable {
+
+    // MARK: Initialization
+
+    /// Instantiates an asynchronous queue.
+    /// - Parameter priority: The priority of the tasks added to the asynchronous queue.
+    public init(priority: TaskPriority = .medium) {
+        var capturedTaskStreamContinuation: AsyncStream<@Sendable () async -> Void>.Continuation? = nil
+        let taskStream = AsyncStream<@Sendable () async -> Void> { continuation in
+            capturedTaskStreamContinuation = continuation
+        }
+        taskStreamContinuation = capturedTaskStreamContinuation
+
+        streamTask = Task.detached(priority: priority) {
+            for await task in taskStream {
+                await task()
+            }
+        }
+    }
+
+    deinit {
+        taskStreamContinuation.finish()
+    }
+
+    // MARK: Public
+
+    /// Schedules an asynchronous task for execution and immediately returns.
+    /// The schedueled task will not execute until all prior tasks have completed.
+    /// - Parameter task: The task to enqueue.
+    public func async(_ task: @escaping @Sendable () async -> Void) {
+        taskStreamContinuation.yield(task)
+    }
+
+    /// Schedules an asynchronous throwing task and returns after the task is complete.
+    /// The schedueled task will not execute until all prior tasks have completed.
+    /// - Parameter task: The task to enqueue.
+    /// - Returns: The value returned from the enqueued task.
+    public func await<T>(_ task: @escaping @Sendable () async -> T) async -> T {
+        await withUnsafeContinuation { continuation in
+            taskStreamContinuation.yield {
+                continuation.resume(returning: await task())
+            }
+        }
+    }
+
+    /// Schedules an asynchronous task and returns after the task is complete.
+    /// The schedueled task will not execute until all prior tasks have completed.
+    /// - Parameter task: The task to enqueue.
+    /// - Returns: The value returned from the enqueued task.
+    public func await<T>(_ task: @escaping @Sendable () async throws -> T) async throws -> T {
+        try await withUnsafeThrowingContinuation { continuation in
+            taskStreamContinuation.yield {
+                do {
+                    continuation.resume(returning: try await task())
+                } catch {
+                    continuation.resume(throwing: error)
+                }
+            }
+        }
+    }
+
+    // MARK: Private
+
+    private let streamTask: Task<Void, Never>
+    private let taskStreamContinuation: AsyncStream<@Sendable () async -> Void>.Continuation!
+}
--- a/Sources/AsyncQueue/AsyncQueue.swift
+++ b/Sources/AsyncQueue/AsyncQueue.swift
@@ -32,7 +32,7 @@ public final class AsyncQueue: Sendable {
         let taskStream = AsyncStream<@Sendable () async -> Void> { continuation in
             capturedTaskStreamContinuation = continuation
         }
-        guard let capturedTaskStreamContinuation = capturedTaskStreamContinuation else {
+        guard let capturedTaskStreamContinuation else {
             fatalError("Continuation not captured during stream creation!")
         }
         taskStreamContinuation = capturedTaskStreamContinuation
--- a/Sources/AsyncQueue/AsyncQueue.swift
+++ b/Sources/AsyncQueue/AsyncQueue.swift
@@ -32,10 +32,9 @@ public final class AsyncQueue: Sendable {
         let taskStream = AsyncStream<@Sendable () async -> Void> { continuation in
             capturedTaskStreamContinuation = continuation
         }
-        guard let capturedTaskStreamContinuation else {
-            fatalError("Continuation not captured during stream creation!")
-        }
-        taskStreamContinuation = capturedTaskStreamContinuation
+        // Continuation will be captured during stream creation, so it is safe to force unwrap here.
+        // If this force-unwrap fails, something is fundamentally broken in the Swift runtime.
+        taskStreamContinuation = capturedTaskStreamContinuation!
 
         streamTask = Task.detached(priority: priority) {
             for await task in taskStream {
--- a/Sources/AsyncQueue/AsyncQueue.swift
+++ b/Sources/AsyncQueue/AsyncQueue.swift
@@ -20,7 +20,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-/// A queue that enables ordered sending of events from synchronous to asynchronous code
+/// A queue that enables sending FIFO-ordered tasks from synchronous to asynchronous contexts
 public final class AsyncQueue: Sendable {
 
     // MARK: Initialization
--- a/Sources/AsyncQueue/AsyncQueue.swift
+++ b/Sources/AsyncQueue/AsyncQueue.swift
@@ -26,7 +26,7 @@ public final class AsyncQueue: Sendable {
     // MARK: Initialization
 
     /// Instantiates an asynchronous queue.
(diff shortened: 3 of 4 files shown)


Sources/AsyncQueue/AsyncQueue.swift
```diff
@@ -20,4 +20,70 @@
...
+                    continuation.resume(throwing: error)
+                }
+            }
+        }
+    }
+
+    // MARK: Private
+
+    private let streamTask: Task<Void, Never>
+    private let taskStreamContinuation: AsyncStream<@Sendable () async -> Void>.Continuation!
```
dfed, line 88: Not ideal that we're force-unwrapping here. I originally made this type optional, but figured we really do want to fail LOUDLY if somehow we don't have a `taskStreamContinuation` and are attempting to append tasks to the queue.

Scripts/build.swift
```diff
@@ -174,6 +174,9 @@ for rawPlatform in rawPlatforms {
     if platform.shouldTest {
         xcodeBuildArguments.append("test")
     }
+    xcodeBuildArguments.append("-test-iterations")
+    xcodeBuildArguments.append("100")
```
dfed, line 178: This is fun. I haven't used these flags before but they enable us to run our tests _a lot_. To help limit our exposure to flakey tests I'm running every test 100 times.

Scripts/build.swift
```diff
@@ -174,6 +174,9 @@ for rawPlatform in rawPlatforms {
...
+    xcodeBuildArguments.append("-run-tests-until-failure")
```
dfed, line 179: This flag stops running tests once they fail. It respects the iterations limit above.

Tests/AsyncQueueTests/AsyncQueueTests.swift
```diff
@@ -26,8 +26,149 @@ import XCTest
...
+            }
+
+            await systemUnderTest.await {
+                let count = await counter.count
+                XCTAssertEqual(count, iteration)
+            }
+        }
+    }
+
+    func test_await_canReturn() async {
```
dfed, line 125: This test and `test_await_canThrow` below are a little silly but I found them helpful for ensuring that the capability exists as I iterated on the queue's API.

Tests/AsyncQueueTests/AsyncQueueTests.swift
```diff
@@ -26,8 +26,149 @@ import XCTest
 
 final class AsyncQueueTests: XCTestCase {
```
dfed, line 27: As you read through the below tests: please be thinking of cases I haven't thought of!

README.md
```diff
@@ -1,5 +1,31 @@
...
-A queue that enables ordered sending of events from synchronous to asynchronous code
+[![Swift Package Manager compatible](https://img.shields.io/badge/SPM-compatible-4BC51D.svg?style=flat)](https://github.com/apple/swift-package-manager)
+[![codecov](https://codecov.io/gh/dfed/swift-async-queue/branch/main/graph/badge.svg?token=nZBHcZZ63F)](https://codecov.io/gh/dfed/swift-async-queue)
+[![Version](https://img.shields.io/cocoapods/v/swift-async-queue.svg)](https://cocoapods.org/pods/swift-async-queue)
+[![License](https://img.shields.io/cocoapods/l/swift-async-queue.svg)](https://cocoapods.org/pods/swift-async-queue)
+[![Platform](https://img.shields.io/cocoapods/p/swift-async-queue.svg)](https://cocoapods.org/pods/swift-async-queue)
+
+A queue that enables ordered sending of events from synchronous to asynchronous code.
+
+## Usage
```
dfed, line 10: This is pretty freaking bare-bones but gotta start somewhere.

Scripts/build.swift
Gabardone: This is a new codebase, do we need to go overkill like that? 😅

Sources/AsyncQueue/AsyncQueue.swift
```diff
@@ -20,4 +20,70 @@
...
 // SOFTWARE.
 
-final class AsyncQueue {}
+/// A queue that enables ordered sending of events from synchronous to asynchronous code
+public final class AsyncQueue: Sendable {
+
+    // MARK: Initialization
+
+    /// Instantiates an asynchronous queue.
+    /// - Parameter priority: The priority of the tasks added to the asynchronous queue.
```
Gabardone, line 29:
```suggestion
    /// - Parameter priority: The baseline priority of the tasks added to the asynchronous queue.
```
I like it better if we hint that priority may go up if the system feels like it.

Sources/AsyncQueue/AsyncQueue.swift
Gabardone: Cane we do the failing at the init method so we can initialize this with a non-optional?

Scripts/build.swift
dfed: Given that we're testing asynchronous code that is potentially quite sensitive to timing issues, I think it'll be helpful!

I found some issues in my earlier implementations by running tests 100 times locally. I figured it'd be good to get CI doing the same.

Sources/AsyncQueue/AsyncQueue.swift
```diff
@@ -20,4 +20,70 @@
...
+    public init(priority: TaskPriority = .medium) {
+        var capturedTaskStreamContinuation: AsyncStream<@Sendable () async -> Void>.Continuation? = nil
+        let taskStream = AsyncStream<@Sendable () async -> Void> { continuation in
+            capturedTaskStreamContinuation = continuation
+        }
+        taskStreamContinuation = capturedTaskStreamContinuation
```
Gabardone, line 35:
```suggestion
        taskStreamContinuation = capturedTaskStreamContinuation!
```
I can't think of any case where this ends up `nil` and we shouldn't blow up anyway.

And deal with any spurious optionality already.

Sources/AsyncQueue/AsyncQueue.swift
dfed: Fair. If you `await` an enqueued task from a high-priority queue then yeah the priority may increase.

Apple's documentation on `Task.detached(...)` just says:
> The priority of the task.

We can do better!

Sources/AsyncQueue/AsyncQueue.swift
```diff
@@ -20,4 +20,70 @@
...
 
-final class AsyncQueue {}
+/// A queue that enables ordered sending of events from synchronous to asynchronous code
+public final class AsyncQueue: Sendable {
+
+    // MARK: Initialization
+
+    /// Instantiates an asynchronous queue.
+    /// - Parameter priority: The priority of the tasks added to the asynchronous queue.
+    public init(priority: TaskPriority = .medium) {
```
Gabardone, line 30:
```suggestion
    public init(priority: TaskPriority? = nil) {
```
Let's follow the Swift library convention for using `TaskPriority` where `nil` means we don't care and let the system figure it out. May need to tweak some of the other properties as well.

Sources/AsyncQueue/AsyncQueue.swift
dfed: This is preferable to making `taskStreamContinuation` implicitly unwrapped because it's more explicit? I went back and forth on this one.

I do agree that at this line the `capturedTaskStreamContinuation` should exist, and it'd be reasonable to blow up the world if it didn't.

Sources/AsyncQueue/AsyncQueue.swift
dfed: Good call!

Tests/AsyncQueueTests/AsyncQueueTests.swift
```diff
@@ -26,8 +26,149 @@ import XCTest
...
+
+    override func tearDown() async throws {
+        try await super.tearDown()
+
+        await systemUnderTest.await { /* Drain the queue */ }
+    }
+
+    // MARK: Behavior Tests
+
+    func test_async_sendsEventsInOrder() {
```
Gabardone, line 45: Sorry to be the nitpickiest guy that has ever nitpicked, but _snake case?_

Gabardone (commented):
See comments but overall looks like a nice thing to have and hopefully we don't have to complicate it much beyond this (at least until such time as Swift or its libraries adds better support for these use cases).

Sources/AsyncQueue/AsyncQueue.swift
dfed: We can, but that would force every call-site to force unwrap, which seems pretty gnarly. Since `taskStreamContinuation` should exist (and it not existing would mean that the world is very broken), I think it's reasonable to crash if it somehow does not exist.

Tests/AsyncQueueTests/AsyncQueueTests.swift
dfed: 😅 yeah my test-naming style is: `test_methodBeingTested_whatIAmTesting`. That format helps me make my test cases more targeted.

Scripts/build.swift
Gabardone: I'm just wondering if it would make more sense to have the tests themselves run something 100 times or this overkill approach where we run all of them 100 times.

At the current code volume it's a distinction without a difference so I'm cool either way.

Sources/AsyncQueue/AsyncQueue.swift
Gabardone: I mean we can do a guard statement or something cleaner than just `!`, what I prefer is to be done with this optionality as soon as possible since it is an artifact of the fact that the Swift compiler can't tell the value will always be set, not of the logic.

Tests/AsyncQueueTests/AsyncQueueTests.swift
Gabardone: I respect that, don't think I've ever had to develop a test naming scheme but I can see it being a good idea.

Sources/AsyncQueue/AsyncQueue.swift
Gabardone: Oh, I meant failing in a very general case. Crashing is fine for something like this where the logic means we would _never_ expect a `nil` value.

Gabardone: The point was to remove the exclamation mark from here (together with the changes in `init` proposed).

Scripts/build.swift
dfed: Agreed we don't need to defend against that future problem just yet. We’ll Get There™. Maybe 🙂

Sources/AsyncQueue/AsyncQueue.swift
dfed: How's this look? 6e0519c

Sources/AsyncQueue/AsyncQueue.swift
```diff
@@ -32,6 +32,9 @@ public final class AsyncQueue: Sendable {
         let taskStream = AsyncStream<@Sendable () async -> Void> { continuation in
             capturedTaskStreamContinuation = continuation
         }
+        guard let capturedTaskStreamContinuation = capturedTaskStreamContinuation else {
```
Gabardone, line 35:
```suggestion
        guard let capturedTaskStreamContinuation else {
```
Are we good with Swift 5.7 or do you need to support older versions of the language?

Gabardone (approved):
See last comment but it's good to merge either way.

Sources/AsyncQueue/AsyncQueue.swift
dfed: I keep forgetting this landed!! Yeah good call.

In most repos I write I support back to Xcode 11 in order to get simulators to test on. But, given that Swift concurrency doesn't go back that far, meh. Let's stick to Xcode 14+

Sources/AsyncQueue/AsyncQueue.swift
dfed: This is silly, but I'm going to use a force-unwrap and a comment instead of a `fatalError`. I want 100% coverage 😅

Gabardone (approved):
I think we ought to merge this already and then keep working on the other PR?

dfed:
> I think we ought to merge this already and then keep working on the other PR?

Yup. Was giving more folk a chance to weigh in, but this repo is very beta (first published version will be `0.0.1`!) and I'm happy to make breaking API changes for a bit.

Folk who come across this PR: please feel free to leave a post-merge review. I will respond, and will happily open up new PRs to address feedback. Given how early we are in the versioning, post-merge naming feedback is still appreciated!

Sources/AsyncQueue/AsyncQueue.swift
```diff
@@ -20,4 +20,72 @@
...
+    }
+
+    deinit {
+        taskStreamContinuation.finish()
+    }
+
+    // MARK: Public
+
+    /// Schedules an asynchronous task for execution and immediately returns.
+    /// The schedueled task will not execute until all prior tasks have completed.
```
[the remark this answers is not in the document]
bachand: Same comment for other documentation comments in this file.

Tests/AsyncQueueTests/AsyncQueueTests.swift
```diff
@@ -26,8 +26,202 @@ import XCTest
...
+            }
+            await counter.incrementAndExpectCount(equals: 1)
+            systemUnderTest.async {
+                await counter.incrementAndExpectCount(equals: 3)
+            }
+        }
+        await systemUnderTest.await { /* Drain the queue */ }
+    }
+
+    func test_async_retainsReceiverUntilFlushed() async {
```
bachand, line 82: What is the receiver in this context?

Tests/AsyncQueueTests/AsyncQueueTests.swift
```diff
@@ -26,8 +26,202 @@ import XCTest
...
+        }
+        // Nil out our reference to the queue to show that the enqueued tasks will still complete
+        systemUnderTest = nil
+        // Signal the semaphore to unlock the remaining enqueued tasks.
+        await semaphore.signal()
+
+        await waitForExpectations(timeout: 1.0)
+    }
+
+    func test_async_doesNotRetainTaskAfterExecution() async {
```
bachand, line 105: Nice test!

Tests/AsyncQueueTests/AsyncQueueTests.swift
```diff
@@ -26,8 +26,202 @@ import XCTest
...
+        let counter = Counter()
+        for iteration in 1...1_000 {
+            systemUnderTest.async {
+                await counter.incrementAndExpectCount(equals: iteration)
+            }
+
+            guard iteration % 25 == 0 else {
+                // Keep sending async events to the queue.
+                continue
+            }
```
bachand, line 143: I don't follow why this code exists in the test. Can you help me understand?

Tests/AsyncQueueTests/AsyncQueueTests.swift
```diff
@@ -26,8 +26,202 @@ import XCTest
...
+    // MARK: Private
+
+    private var systemUnderTest = AsyncQueue()
+
+    // MARK: - Counter
+
+    private actor Counter {
+        func incrementAndExpectCount(equals expectedCount: Int) {
+            increment()
+            XCTAssertEqual(expectedCount, count)
```
bachand, line 180: Nit: would be nice to pass through the file and line number so any XCTest failure is easier to debug.

bachand (commented):
Nice!

README.md
```diff
@@ -1,5 +1,31 @@
 # swift-async-queue
-A queue that enables ordered sending of events from synchronous to asynchronous code
+[![Swift Package Manager compatible](https://img.shields.io/badge/SPM-compatible-4BC51D.svg?style=flat)](https://github.com/apple/swift-package-manager)
```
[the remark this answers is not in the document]
dfed: Yup. I need to publish the pod before those badges will work. I'll remove them for now: 51e98a3

Sources/AsyncQueue/AsyncQueue.swift
```diff
@@ -20,4 +20,72 @@
...
+    public init(priority: TaskPriority? = nil) {
+        var capturedTaskStreamContinuation: AsyncStream<@Sendable () async -> Void>.Continuation? = nil
+        let taskStream = AsyncStream<@Sendable () async -> Void> { continuation in
+            capturedTaskStreamContinuation = continuation
+        }
+        // Continuation will be captured during stream creation, so it is safe to force unwrap here.
+        // If this force-unwrap fails, something is fundamentally broken in the Swift runtime.
+        taskStreamContinuation = capturedTaskStreamContinuation!
+
+        streamTask = Task.detached(priority: priority) {
```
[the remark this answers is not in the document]
dfed: Exactly!

Sources/AsyncQueue/AsyncQueue.swift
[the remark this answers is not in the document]
dfed: Good catch! 2bd88a1

Tests/AsyncQueueTests/AsyncQueueTests.swift
dfed: The receiver is the `systemUnderTest`, i.e. the async queue.

Tests/AsyncQueueTests/AsyncQueueTests.swift
dfed: This was a fun one to write 😄

Tests/AsyncQueueTests/AsyncQueueTests.swift
dfed: I wanted some way to interleave multiple `async` events with each `await` event that was deterministic. We could `await` after every `async`, but because the `await` method requires waiting for the work to finish I felt like the test wouldn't necessarily catch an issue where `await` wouldn't wait for every enqueued event but would just wait for the first enqueued event...

Short version: this `guard` could be removed and it'd probably be fine. But I kinda like emitting multiple `async` events for every `await` check.

Tests/AsyncQueueTests/AsyncQueueTests.swift
dfed: b34e05a

Tests/AsyncQueueTests/AsyncQueueTests.swift
bachand: Ah got it. I understand now 👍

PR #3, opened 2022-11-24
Rename AsyncQueue -> FIFOQueue. Create ActorQueue

dfed:
I realized that our `AsyncQueue` is really a `FIFOQueue`, and that consumers may desire a queue which maintains task execution order when interacting with an `actor` but allows for reentrancy and re-ordering of tasks due to suspension.

This realization led to the creation of an `ActorQueue`. This queue is purpose-built to send ordered, asynchronous work to `actor` instances from a non-isolated context/

Reviewers should feel free to question the type and method names within this PR. I could use another brain here.


Tests/AsyncQueueTests/FIFOQueueTests.swift
```diff
@@ -172,58 +172,5 @@ final class AsyncQueueTests: XCTestCase {
...
-        func increment() {
-            count += 1
-        }
-
-        var count = 0
-    }
-
-    // MARK: - Semaphore
-
-    private actor Semaphore {
```
dfed, line 194: This `Semaphore` implementation was moved to its own file without changes.

Tests/AsyncQueueTests/FIFOQueueTests.swift
```diff
@@ -172,58 +172,5 @@ final class AsyncQueueTests: XCTestCase {
 
     // MARK: Private
 
-    private var systemUnderTest = AsyncQueue()
-
-    // MARK: - Counter
-
-    private actor Counter {
```
dfed, line 179: This `Counter` implementation was moved to its own file without changes

Tests/AsyncQueueTests/FIFOQueueTests.swift
```diff
@@ -24,34 +24,29 @@ import XCTest
...
 
     // MARK: XCTestCase
 
     override func setUp() async throws {
         try await super.setUp()
 
-        systemUnderTest = AsyncQueue()
-    }
-
-    override func tearDown() async throws {
```
dfed, line 37: this `tearDown` method did not have any effect on whether tests would pass or not, so I deleted it

Tests/AsyncQueueTests/FIFOQueueTests.swift
```diff
@@ -24,34 +24,29 @@ import XCTest
...
 
-    func test_async_sendsEventsInOrder() {
+    func test_async_sendsEventsInOrder() async {
         let counter = Counter()
         for iteration in 1...1_000 {
             systemUnderTest.async {
                 await counter.incrementAndExpectCount(equals: iteration)
             }
         }
+        await systemUnderTest.await { /* Drain the queue */ }
```
dfed, line 46: We must drain the queue within the body of the test method in order to catch test failures. Test failures that occur after the test method has exited do not fail the test run.

Sources/AsyncQueue/Semaphore.swift
```diff
@@ -0,0 +1,57 @@
...
+/// A thread-safe semaphore implementation.
+actor Semaphore {
+    func wait() async {
+        count -= 1
+        guard count < 0 else {
+            // We don't need to wait because count is greater than or equal to zero.
+            return
+        }
+
+        await withCheckedContinuation { continuation in
```
dfed, line 32:
```suggestion
        await withUnsafeContinuation { continuation in
```
We have enough unit tests here that we don't need to utilize a checked continuation.

README.md
```diff
@@ -5,26 +5,77 @@
...
+        Work enqueued after this task will wait for this task to complete.
+        */
+    }
+}
 ```
 
-### Awaiting work from an asynchronous context
+### Sending ordered asynchronous tasks to Actors
+
+Use an `ActorQueue` queue to send ordered asynchronous tasks from a nonisolated context to an `actor` instance. Tasks sent to one of these queues are guaranteed to begin _but not end_ executing in the order in which they are enqueued.
```
Gabardone, line 60: I think going a bit more into detail would be good.
```suggestion
Use an `ActorQueue` queue to send ordered asynchronous tasks from a nonisolated context to an `actor` instance. Tasks sent to one of these queues are guaranteed to begin executing by the actor in the order in which they are enqueued. Whether those tasks will also end in the same order depends on the actor's implementation, as it may be using concurrency internally itself.
```

Gabardone: See later comments though. This is a gnarly thing to define properly.

README.md
```diff
@@ -5,26 +5,77 @@
...
 
 ```swift
-await asyncQueue.await { /* throw-able, return-able, awaitable context that executes after all other enqueued work is completed */ }
+let queue = ActorQueue()
+queue.async {
+    /*
+    `async` context that executes after all other enqueued work has begun executing.
```
Gabardone, line 66: Is it possible to call non-async actor methods synchronously from the outside, by setting us in the same execution context or something like that?

Actors have access to their own serial executors so it may work out to have some kind of actor extension that runs non-serial code through the executor. Although Swift doesn't let us extend a protocol so we'd need to put some ornamentation around such an idea.

I'll need to discuss this further, I'm sure you've been giving this a lot of thought as well.

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -0,0 +1,116 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+/// A queue that executes asynchronous tasks enqueued from a nonisolated context.
+/// Tasks are guaranteed to begin executing in the order in which they are enqueued. However, if a task suspends it will allow tasks that were enqueued to begin executing.
+/// Asynchronous tasks sent to this queue execute as they would in an `actor` type, allowing for re-entrancy and non-FIFO behavior when an individual task suspends.
+/// - Warning: Execution order is not guaranteed unless the enqueued tasks interact with a single `actor` instance.
```
Gabardone, line 26: It may make sense to attach the actor itself to the queue? Since other ways of doing this seem to hit against the limitations of the Swift type system...

I don't foresee a sensible use case for this where the same queue is used for several different actors.

dfed:
Force push was rebasing `develop`

README.md
dfed: > Is it possible to call non-async actor methods synchronously from the outside, by setting us in the same execution context or something like that?

Yes, if you use [`@_inheritActorContext`](https://github.com/apple/swift/blob/1f3e159cfe518fbb723230a4e8f562248d34d863/docs/ReferenceGuides/UnderscoredAttributes.md#_inheritactorcontext). But given the `Usage of these attributes outside of the Swift monorepo is STRONGLY DISCOURAGED` warning at the top of that link, I do not plan on using underscored attributes in this repo.

> Actors have access to their own serial executors so it may work out to have some kind of actor extension that runs non-serial code through the executor.

This would be ideal. I haven't found a supported way to do this though.

> Although Swift doesn't let us extend a protocol so we'd need to put some ornamentation around such an idea.

Indeed we'd need to at least be able to store the `taskStreamContinuation`, and we can't do that on a protocol extension. Worse yet, `actor`s do not support inheritance, so we're quite low on options here.

Sources/AsyncQueue/ActorQueue.swift
dfed: I thought about trying to associate a single actor with a queue, but fell short coming up with a good way to do it. As-written, I expect folk will utilize an `ActorQueue` as follows:

```
actor MyActor {

    nonisolated
    func send(input: Input) {
        queue.async {
            await self.process(input: input)
        }
    }

    private func process(input: Input) {
        // Process the ordered input on the actor queue
    }

    private let queue = ActorQueue()

}
```

This isn't a whole lot of boilerplate, and I haven't come up with something cleaner.

I'm not terribly worried about folk mis-using this API assuming we get the documentation right. Folk coming to this repo are likely trying to solve a very specific problem and will use this code as intended (again, assuming the docs are good).

That said, I'm open to ideas. I've been unable to come up with anything better thus far than what we've got here, but that doesn't mean there aren't better options!

dfed: For what it's worth, I tried a ton of different shenanigans to get rid of this warning but came up short. What I need is a way to signal the semaphore after we've suspended the current context's execution to `await` the task. Basically I need something like:

```
// Signal the semaphore once we've begun waiting for the `task`.
deferUntilSuspension {
    await semaphore.signal()
}
await task()
```

But that isn't an existing language feature!

README.md
dfed: We're basically talking about FIFE (first-in-first-executing) contract, with ordering guarantees up until the first suspension point. So maybe we define this as:
```suggestion
Use an `ActorQueue` queue to send ordered asynchronous tasks from a nonisolated context to an `actor` instance. Tasks sent to one of these queues are guaranteed to begin executing in the order in which they are enqueued. Ordering of execution is guaranteed up until the first [suspension point](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html#ID639) within the called `actor` code.
```

Gabardone: Ok that is probably better, with concurrent logic around we always get to the point where we just 🤷🏽‍♂️ and hope what the caller is doing very soon.

I'd like to experiment with sending synchronous code to run, but I don't think Swift will let us as currently configured (maybe the actor can run a serial block block itself but how to declare that in a way that will always work I'm not sure).

Sources/AsyncQueue/ActorQueue.swift
Gabardone: A possible alternative would be to store the actor in the queue class and then also pass it to the execution blocks as a parameter. Decouples things a bit more than the raw blocks here.

README.md
dfed: > I'd like to experiment with sending synchronous code to run, but I don't think Swift will let us as currently configured

This is doable! I've written + deleted this test a few times in the process of discovering + trying to delete the warning I put on this type. Looks like this:
```
func test_async_sendsSynchronousEventsInOrder() async {
    final class SynchronousCounter: @unchecked Sendable {
        func incrementAndExpectCount(equals expectedCount: Int) {
            increment()
            XCTAssertEqual(expectedCount, count)
        }

        func increment() {
            count += 1
        }

        var count: Int {
            set {
                dispatchQueue.async { self.unsafeCount = newValue }
            }
            get {
                dispatchQueue.sync { unsafeCount }
            }
        }
        var unsafeCount = 0
        let dispatchQueue = DispatchQueue(label: #function)
    }
    let counter = SynchronousCounter()
    for iteration in 1...1_000 {
        systemUnderTest.async {
            counter.incrementAndExpectCount(equals: iteration)
        }
    }
    await systemUnderTest.await { /* Drain the queue */ }
}
```

dfed: Note that in the above code we don't really need the `dispatchQueue` barrier – I added it only to prevent a crash in case code were unexpectedly executed concurrently.

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -99,18 +78,90 @@ public final class ActorQueue: Sendable {
...
+            _ task: TaskType,
+            afterSignaling semaphore: Semaphore)
+        {
+            // Utilize the serial (but not FIFO) Actor context to execute the task without requiring the calling method to wait for the task to complete.
+            Task {
+                switch task {
+                case let .sync(task):
+                    task()
+                    // Synchronous tasks can not re-enter this queue, so it is safe to wait until the task completes prior to signaling the semaphore.
+                    await semaphore.signal()
```
dfed, line 157: `test_async_sendsSynchronousEventsInOrder` fails unless I `signal` the semaphore after the `task()` completes. I don't fully understand why just yet.

This task is executing within a serial actor context, so the `task` should complete within the same serial context as the semaphore signal. My best guess at an explanation is that signaling the semaphore requires suspension, and so ordering post-suspension to interact with the semaphore has no guaranteed order. However, I don't know how that hypothesis jives with the fact that the `.async`  case below works (and has worked on over 15mil repetitions).

I would love to better understand this behavior. Insights are appreciated!

As an aside: I am very much open to the feedback that this `.sync` case should exist on this queue at all. I added it to start a discussion around suspension, and am not convinced that this API should belong on the queue.

Sources/AsyncQueue/ActorQueue.swift
dfed: I'm not convinced that would work, since any suspension to signal the semaphore would potentially result in the actor methods executing in an unexpected order 🤔. See https://github.com/dfed/swift-async-queue/pull/3/commits/0565678b069a5c6d195113e261ecf449d626cbfa#r1033680644

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -0,0 +1,167 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+/// A queue that executes asynchronous tasks enqueued from a nonisolated context.
```
bachand, line 23: It would be nice I think for the doc comment to include sample code like in https://github.com/dfed/swift-async-queue/pull/3/files#r1033055297 to show how this class is meant to be used.

Sources/AsyncQueue/ActorQueue.swift
bachand: > As an aside: I am very much open to the feedback that this `.sync` case should exist on this queue at all. I added it to start a discussion around suspension, and am not convinced that this API should belong on the queue.

Based on the intended usage as show in https://github.com/dfed/swift-async-queue/pull/3/files#r1033055297 it seems like there would never be a need for the synchronous methods, right?

bachand: > `test_async_sendsSynchronousEventsInOrder` fails unless I `signal` the semaphore after the `task()` completes. I don't fully understand why just yet.
> 
> This task is executing within a serial actor context, so the `task` should complete within the same serial context as the semaphore signal. My best guess at an explanation is that signaling the semaphore requires suspension, and so ordering post-suspension to interact with the semaphore has no guaranteed order. However, I don't know how that hypothesis jives with the fact that the `.async`  case below works (and has worked on over 15mil repetitions).
> 
> I would love to better understand this behavior. Insights are appreciated!

I feel like the behavior is due to this code. https://github.com/dfed/swift-async-queue/blob/17894cc8e356de009fe93d3855715d47ab0b4794/Sources/AsyncQueue/ActorQueue.swift#L42-L47

If we signal the semaphore before executing the task then that code can begin processing the next task before the prior task has been completed. And the next task can be processed in parallel and is thus racing the previous task.

bachand: I am also a bit confused initially why the `.async` case passes when we signal before executing the task.

bachand: I think the `.async` case works because the task closure in that case is async itself, so it gets ordering from the actor. In the `.sync` case the task closure is not async so there is nothing enforcing the order of execution. I agree this is pretty trippy to think through 😵

Sources/AsyncQueue/FIFOQueue.swift
```diff
@@ -20,8 +20,10 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-/// A queue that enables sending FIFO-ordered tasks from synchronous to asynchronous contexts
-public final class AsyncQueue: Sendable {
+/// A queue that executes asynchronous tasks enqueued from a nonisolated context in FIFO order.
```
bachand, line 23: As before I think it would be helpful to have a code example showing how this class is intended to be used.

Sources/AsyncQueue/Semaphore.swift
```diff
@@ -0,0 +1,57 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+/// A thread-safe semaphore implementation.
+actor Semaphore {
```
bachand, line 24: Naive and potentially lazy question: why not use (or extend) `DispatchSemaphore`?

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -0,0 +1,91 @@
...
+
+        for signal in signals {
+            await signal.value
+        }
+
+        await self.systemUnderTest.signal()
+
+        for wait in waits {
+            await wait.value
+        }
```
bachand, line 76: We want to execute the waits before the signals right? In the wait is where we increment. It seems like we need to execute the waits first for L64 (`XCTAssertEqual(0, count)`) to be a meaningful test.

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -0,0 +1,91 @@
...
+
+        let count = await counter.count
+        XCTAssertEqual(iterationCount, count)
+    }
+
+    func test_wait_doesNotSuspendIfSignalCalledFirst() async {
+        await systemUnderTest.signal()
+        await systemUnderTest.wait()
+        // If the test doesn't hang forever, we've succeeded!
```
bachand, line 85: The test name says the we are verifying we don't suspend whereas the test itself is verifying that we eventually stop being suspended. I think we should bring the test name and the test content in sync.

Tests/AsyncQueueTests/ActorQueueTests.swift
```diff
@@ -0,0 +1,184 @@
...
+        let counter = SynchronousCounter()
+        for iteration in 1...1_000 {
+            systemUnderTest.async {
+                counter.incrementAndExpectCount(equals: iteration)
+            }
+        }
+        await systemUnderTest.await { /* Drain the queue */ }
+    }
+
+    func test_async_startsExecutionOfNextTaskAfterSuspension() async {
```
bachand, line 80: I could be missing something but would this test also pass with a `FIFOQueue`? I'm trying to think if there's a more targeted way to verify/show that the next task starts to execute after the first task suspends. It may be that we don't need 1000 tasks for this type of test, and just want a couple?

Tests/AsyncQueueTests/ActorQueueTests.swift
```diff
@@ -0,0 +1,184 @@
...
+        referenceHolder.reference = nil
+        XCTAssertNotNil(weakReference)
+        // Allow the enqueued task to complete.
+        await asyncSemaphore.signal()
+        // Make sure the task has completed.
+        waitForExpectations(timeout: 1.0)
+        // Wait a runloop to ensure the previously enqueued task has had time to deallocate.
+        DispatchQueue.main.async {
+            XCTAssertNil(weakReference)
+        }
```
bachand, line 140: We didn't need this code to wait a runloop in the `FIFOQueue` tests. Why do we need it here?

bachand: Also, by dispatching async on the main queue the assertion runs after this method completes. If that's the case and I'm not missing something it's something we should avoid. Can we use an `@autoreleasepool` instead?

bachand (commented):
Nice work @dfed . My main feedback (as I noted in a couple of comments) is that it would be nice to have more clarity on how these different types are intended to be used. I feel like writing that out will make the library more approachable, and may guide future API iteration.

Sources/AsyncQueue/ActorQueue.swift
dfed: > it seems like there would never be a need for the synchronous methods, right?

There's not a need, no. But it could be a nice-to-have, if confusing feature. I'll remove the `.sync` option on my next pass.

dfed: > If we signal the semaphore before executing the task then that code can begin processing the next task before the prior task has been completed. And the next task can be processed in parallel and is thus racing the previous task.

Yes, but the `executor` in the code you linked is an actor, and we're already processing code within the actor so it shouldn't be possible for the next task to begin execution on the actor context...

And of course as I typed that out I realized it's not possible for the next task to begin execution on the actor context _until the actor context suspends_, and we are suspending when we interact with the semaphore.

dfed: > I think the `.async` case works because the task closure in that case is async itself, so it gets ordering from the actor. In the `.sync` case the task closure is not async so there is nothing enforcing the order of execution. I agree this is pretty trippy to think through 😵

This is my understanding as well, though I wish I better understood what was happening under the hood. It feels like the code I've written works by accident rather than intention. But I've run the async ordering test case 20+million times and it has never failed...

README.md
```diff
@@ -5,26 +5,77 @@
...
+    Work enqueued after this task will wait for this task to complete or suspend.
+    */
+}
+Task {
+    await queue.await {
+        /*
+        `async` context that can return a value or throw an error.
+        Executes after all other enqueued work has completed or suspended.
```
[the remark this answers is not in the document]
dfed: yup. I wrote this description before I'd discovered that the synchronous test failed.

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -0,0 +1,167 @@
...
+/// Tasks are guaranteed to begin executing in the order in which they are enqueued. However, if a task suspends it will allow tasks that were enqueued to begin executing.
+/// Asynchronous tasks sent to this queue execute as they would in an `actor` type, allowing for re-entrancy and non-FIFO behavior when an individual task suspends.
+/// - Warning: Execution order is not guaranteed unless the enqueued tasks interact with a single `actor` instance.
+public final class ActorQueue: Sendable {
```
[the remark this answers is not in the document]
dfed: Good point.
```suggestion
public final class ActorQueue {
```

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -0,0 +1,167 @@
...
+
+    // MARK: Initialization
+
+    /// Instantiates an asynchronous queue.
```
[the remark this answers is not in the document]
dfed: 6fa22de

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -0,0 +1,167 @@
...
+    // MARK: Public
+
+    /// Schedules an asynchronous task for execution and immediately returns.
+    /// The schedueled task will not execute until all prior tasks have completed or suspended.
+    /// - Parameter task: The task to enqueue.
+    public func async(_ task: @escaping @Sendable () async -> Void) {
+        taskStreamContinuation.yield(.async(task))
+    }
+
+    /// Schedules an asynchronous throwing task and returns after the task is complete.
```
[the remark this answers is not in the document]
dfed: 6fa22de

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -0,0 +1,167 @@
...
+    /// - Returns: The value returned from the enqueued task.
+    public func await<T>(_ task: @escaping @Sendable () async -> T) async -> T {
+        await withUnsafeContinuation { continuation in
+            taskStreamContinuation.yield(.async({
+                continuation.resume(returning: await task())
+            }))
+        }
+    }
+
+    /// Schedules an asynchronous task and returns after the task is complete.
```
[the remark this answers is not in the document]
dfed: 6fa22de

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /// Schedules a synchronous task for execution and immediately returns.
+    /// The schedueled task will not execute until all prior tasks have completed or suspended.
+    /// - Parameter task: The task to enqueue.
+    public func async(_ task: @escaping @Sendable () -> Void) {
+        taskStreamContinuation.yield(.sync(task))
+    }
+
+    /// Schedules an throwing task and returns after the task is complete.
```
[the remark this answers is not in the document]
dfed: 6fa22de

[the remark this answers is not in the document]
dfed: 1f656db

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    // MARK: Private
+
+    private let taskStreamContinuation: AsyncStream<TaskType>.Continuation
+
+    // MARK: - TaskType
+
+    enum TaskType {
```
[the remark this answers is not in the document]
dfed: yup, though I'm going to remove `TaskType` I think on the next pass

Sources/AsyncQueue/Semaphore.swift
dfed: I didn't want to block a queue, I just wanted to suspend execution. `DispatchSemaphore` is blocking, whereas this `Semaphore` just suspends execution.

Tests/AsyncQueueTests/ActorQueueTests.swift
dfed: This test deadlocks when the `systemUnderTest` is a `FIFOQueue` instance.

The reason it deadlocks is:
1. A `systemUnderTest.async` task `await`s the semaphore
2. A future `systemUnderTest.async` task signals the semaphore, but that future task never executes because the first task is stuck waiting for the semaphore to unlock

Tests/AsyncQueueTests/SemaphoreTests.swift
dfed: Indeed! This one is a little tricky and I don't love it.

We can't `await wait.value` until we've signaled the semaphore or we'll deadlock. So what I do is spawn 1000 `Task`s that wait for the semaphore assuming that at least one of them will execute before we finish waiting for the signals. Based on code coverage when run locally, this does occur.

I'm open to better ways to write this test, but Swift's concurrency model doesn't give me a great way to say "wait until this task has started" or "wait until this task has suspended", which is what we'd need to make this test more deterministically meaningful.

Tests/AsyncQueueTests/SemaphoreTests.swift
dfed: the `wait` in this test never suspends its own execution. We suspend _the test_ code to enter the semaphore's context but the semaphore itself does not suspend.

Thoughts on improving this test by making `wait` return a discardable boolean result that is `true` whenever the `wait` call resulted in suspension? That would make this test a bit more explicit without complicating call-sites.

Sources/AsyncQueue/ActorQueue.swift
dfed: I'll take a documentation pass in the next couple days. Agreed this documentation needs some strengthening.

Tests/AsyncQueueTests/ActorQueueTests.swift
bachand: Got it. Maybe we can add some comments connecting the dots in the test code for how we are proving that a task starts before the previous task finishes?

dfed: Good call! Will do on the next pass

Sources/AsyncQueue/ActorQueue.swift
bachand: It's certainly going to take me a little while to establish a mental model for Swift concurrency.

Tests/AsyncQueueTests/ActorQueueTests.swift
bachand: If we do need to use an async onto the main queue I think we want to write an expectation that we fulfill after the async code completes, and then wait for the expectation at the end of the test. That ensures that we always execute the `XCTAssertNil` before `test_async_doesNotRetainTaskAfterExecution` returns.

bachand: > I tried writing this test with an `autoreleasepool` those do not work with `async` closures.

Would this approach help us? https://stackoverflow.com/a/68976711/280789 If so this feels a bit more explicit about what we're trying to achieve wrt memory management.

dfed: The issue is less that ARC hasn't finished yet as much as it is that **guaranteeing** that we execute _after_ the block has completed execution is difficult. I don't think an autorelease pool will help us much, especially one wrapped in a `Task` because controlling when a `Task` executes is seriously difficult.

dfed: > If we do need to use an async onto the main queue I think we want to write an expectation that we fulfill after the async code completes, and then wait for the expectation at the end of the test. That ensures that we always execute the `XCTAssertNil` before `test_async_doesNotRetainTaskAfterExecution` returns.

Good point that we might execute the `XCTAssertNil` too late though. That's hopefully an easy fix with an `await` and another expectation.

Tests/AsyncQueueTests/ActorQueueTests.swift
dfed: 583ca33

Tests/AsyncQueueTests/ActorQueueTests.swift
dfed: Simplified dramatically in b313127. The key was that I remembered that the actor queue is still executing tasks atomically _as long as those tasks do not await anything_.

Tests/AsyncQueueTests/SemaphoreTests.swift
bachand: > We can't `await wait.value` until we've signaled the semaphore or we'll deadlock.

👍 

> I'm open to better ways to write this test, but Swift's concurrency model doesn't give me a great way to say "wait until this task has started" or "wait until this task has suspended", which is what we'd need to make this test more deterministically meaningful.

I got into the code and I see what you are saying. The way this test is written I believe that it is _possible_ for it to fail intermittently. I can see the test failing if the signaling tasks were scheduled more frequently than the waiting tasks. That could result in a situation where the counter was incremented before all signaling was complete.

I also am not clear on how to make this test more deterministic 🤔

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -0,0 +1,91 @@
...
+                let count = await counter.count
+                XCTAssertEqual(0, count)
+            })
+        }
+
+        for signal in signals {
+            await signal.value
+        }
+
+        await self.systemUnderTest.signal()
```
bachand, line 72: Can you explain why this signal is standalone and not part of the above loop?

Tests/AsyncQueueTests/SemaphoreTests.swift
dfed: > I also am not clear on how to make this test more deterministic 🤔

This is why I am running all tests 100 times in CI, and using large numbers of repetitions in tests. I'm doing my best to make tests deterministic, but some tests aren't possible to make so. Therefore I rely on repetition to catch issues.

Locally as I develop I run tests 1000 times after I've made a change I'm not sure about just to make sure 😅

Clearly not ideal, but these are the tools I see so I'm using them!

bachand: Thanks for talking this through!

Tests/AsyncQueueTests/SemaphoreTests.swift
bachand: Those comments are helpful. To make sure I understand... the reason the final signal is inlined into the test is because we don't want to run 

```swift
let count = await counter.count
XCTAssertEqual(0, count)
```

ffter the last signal since that assertion is (more?) likely to fail. Does that sound right?

Tests/AsyncQueueTests/SemaphoreTests.swift
dfed: 2863b37a580ee373cec3887af3a9a09be42fd41c

Sources/AsyncQueue/Semaphore.swift
```diff
@@ -0,0 +1,62 @@
...
+    /// Increment the counting semaphore. If the previous value was less than zero, this function resumes a waiting thread before returning.
+    func signal() {
+        count += 1
+        guard !isWaiting else {
+            // Continue waiting.
+            return
+        }
+
+        for continuation in continuations {
+            continuation.resume()
```
bachand, line 50: Asking for my own learning (and totally fine if you don't know). When we call `continuation.resume()` does that mean that we will begin executing the suspended code immediately on the current queue? Or do we signal to the system that the suspended code should resume and the system decides when/where to run it later?

bachand: If you don't know I can try this out in a playground maybe and see.

dfed: I had this same question – thank you for giving me the push to dig in! Turns out [Apple's docs are great here](https://developer.apple.com/documentation/swift/unsafecontinuation/resume()):

> After calling this method, control immediately returns to the caller. The task continues executing when its executor schedules it.

Sounds a lot like "we signal to the system that the suspended code should resume and the system decides when/where to run it later"

Tests/AsyncQueueTests/Utilities/Counter.swift
```diff
@@ -0,0 +1,36 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import XCTest
+
+actor Counter {
```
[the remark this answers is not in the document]
dfed: No more `_inSynchronizationQueue`-suffixed methods! Absolutely lowers the barrier to entry for folk maintaining thread-safe types.

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -0,0 +1,126 @@
...
+        }
+        let executor = CountingExecutor()
+        let iterationCount = 1_000
+
+        for _ in 1...iterationCount {
+            await executor.enqueue(andCount: true) {
+                let didSuspend = await self.systemUnderTest.wait()
+                XCTAssertTrue(didSuspend)
+
+                // Signal without waiting that our prior wait completed.
```
bachand, line 81: Won't we only reach this line of code if the previous wait completes? Maybe I'm not following what "completed" is meant to mean?

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        await executor.execute {
+            // Signal one last time, enabling all of the original `wait` calls to resume.
+            await self.systemUnderTest.signal()
+
+            for _ in 1...iterationCount {
+                // Wait for all enqueued `wait`s to have completed and signaled their completion.
+                await self.systemUnderTest.wait()
+            }
```
bachand, line 108: Is it necessary to loop or would a single `await self.systemUnderTest.wait()` have the same result?

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -0,0 +1,126 @@
...
+                let tasksCompleted = await executor.countedTasksCompleted
+                XCTAssertEqual(iterationCount, tasksCompleted)
+            }
+        }
+    }
+
+    func test_wait_doesNotSuspendIfSignalCalledFirst() async {
+        await systemUnderTest.signal()
+        let didSuspend = await systemUnderTest.wait()
+        XCTAssertFalse(didSuspend)
```
bachand, line 120: 👍

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -0,0 +1,126 @@
...
+    // MARK: Behavior Tests
+
+    func test_wait_suspendsUntilEqualNumberOfSignalCalls() async {
+        actor CountingExecutor {
+            func enqueue(andCount incrementOnCompletion: Bool = false, _ task: @escaping @Sendable () async -> Void) async {
+                await withCheckedContinuation { continuation in
+                    // Re-enter the actor context but don't wait for the result.
+                    Task {
+                        // Now that we're back in the actor context, resume.
+                        continuation.resume()
```
bachand, line 54: I'm having trouble following the continuation -> task dance that we're doing here. Can you help me follow why that's necessary? When `enqueue(…)` starts executing we're already on the actor context so why is it necessary to create a continuation and then create a task to get back on the actor context?

bachand: Overall I think an explanation would be helpful for how the inlined actor improved this test. Based on https://github.com/dfed/swift-async-queue/pull/3#discussion_r1041229725 I see that this is definitely an improvement, though I'd love to better understand the logic/reasoning.

dfed: I'll add some comments in the next pass. Short version: we're using a trick similar to the `ActorQueue` to enable ordered execution of tasks from a single context.

The problems I was solving for:
1. I needed to ensure we are calling `wait()` before `signal()`
2. I needed to not `await` the `wait()` call, since that would have hung the test

In order to do that, I used an `Actor` (which has inherit task ordering) to serialize enqueued tasks and ensure the `wait()` call would suspend because it was executing before we `signal()`'d.

Tests/AsyncQueueTests/SemaphoreTests.swift
dfed: > Won't we only reach this line of code if the previous wait completes? 

Correct.

We use the signal below to signal to the `wait()` on line 107. We need line 107 and after to execute _after_ this task has completed and we've incremented the counter.

Tests/AsyncQueueTests/SemaphoreTests.swift
dfed: We want to `wait()` for every `signal()` on line 83, so that we only continue execution once _all_ of the tasks enqueued on line 77 have completed.

Tests/AsyncQueueTests/SemaphoreTests.swift
dfed: Next pass comments: d1a5560

Sources/AsyncQueue/ActorQueue.swift
dfed: Welp I've been slow here. Going to do a documentation pass today.

dfed: Took a pass here: 744db4a

I put the code in the README rather than the the doc comment since this is a relatively complex example for a doc comment. That said, if you've got examples of code that has built complex code samples into doc comments, please do link me!

There's still room for improvement in the docs, but I think we're in a place where we could land this and improve over time.

dfed: `AsyncStream` is a GREAT example of putting this documentation on the type in the code. I'll give this another pass, though I think we can likely merge as-is and take another docs pass post-merge.

dfed: Took another pass because why not: dca625d

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -0,0 +1,144 @@
...
+
+    private actor ActorExecutor {
+        func suspendUntilStarted(_ task: @escaping @Sendable () async -> Void) async {
+            // Suspend the calling code until our enqueued task starts.
+            await withUnsafeContinuation { continuation in
+                // Utilize the serial (but not FIFO) Actor context to execute the task without requiring the calling method to wait for the task to complete.
+                Task {
+                    // Force this task to execute within the ActorExecutor's context by accessing an ivar on the instance.
+                    // Without this line the task executes on a random context, causing execution order to be nondeterministic.
+                    _ = void
```
dfed, line 132: I've been able to reproduce this issue on tvOS and macOS, but not iOS. I haven't tested watchOS just yet. Looking into whether this is a known issue. Otherwise I'll file a bug.

Sources/AsyncQueue/ActorQueue.swift
dfed: Resolving this thread since it's been idle for a bit and neither of us have come up with good alternatives. Feel free to un-resolve.

Sources/AsyncQueue/FIFOQueue.swift
dfed: I've been reflecting on this, and this queue doesn't really have an "intended" way to be used. It enables executing async tasks from a synchronous context in FIFO order, but unlike the `ActorQueue` there's no real limitations or intended usage here.

dfed:
Apologies for the multiple rounds of revisions here. @bachand @Gabardone this PR is once again ready for another review!

Sources/AsyncQueue/ActorQueue.swift
dfed: With 10ec535 I think I have a much better understanding of how this works.

We are blocking the calling async for-loop until we've entered a `Task` running within an `actor`'s context. When we suspend to interact with another `actor` within the `await task()` line, we enqueue work for that other `actor` in order.

If the `task()` closure is synchronous, the `await` triggers suspension but since we don't have ordering from interacting with another actor, the suspension leads to random ordering.

The reason we were seeing `test_async_sendsSynchronousEventsInOrder` fail unless I `signal`d the semaphore until after the `task()` completed (see the OP on this thread) was because of [this bug](https://github.com/apple/swift/issues/62503) – adding the `_ = void` trick seen in 10ec535 enabled me to get `test_async_sendsSynchronousEventsInOrder` passing when `signal`ing the semaphore _before_ executing the `task()`.

Once this PR lands I think I'll try to enable adding synchronous tasks back to the `ActorQueue` (though I'm still waffling on it). It could be useful to ensure ordering of events enqueued from the isolated context and nonisolated contexts. We’ll Get There™, maybe.

Since we've now got a good handle on what was happening here, I'm resolving this thread.

Tests/AsyncQueueTests/ActorQueueTests.swift
```diff
@@ -0,0 +1,181 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import XCTest
+
+@testable import AsyncQueue
+
+final class ActorQueueTests: XCTestCase {
```
dfed, line 27: When reviewing this file, it's worth comparing these tests to the ones in `FIFOQueueTests`

README.md
dfed: Following up: I don't think our use case for an `ActorQueue` would result in us wanting to inherit the actor context. The entire point of an `ActorQueue` is to bridge the nonisolated context with the actor-isolated context – inheriting an actor context from a nonisolated context doesn't improve the semantics since we're calling into the `ActorQueue` from a context other than the actor's.

Tests/AsyncQueueTests/SemaphoreTests.swift
dfed: Re-closing the loop here: I pulled this test out into https://github.com/dfed/swift-async-queue/pull/5, and was able to get it to deterministically fail when the we `signal` too many times!

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -0,0 +1,144 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+/// A queue that executes asynchronous tasks enqueued from a nonisolated context.
+/// Tasks are guaranteed to begin executing in the order in which they are enqueued. However, if a task suspends it will allow subsequently enqueued tasks to begin executing.
+/// Asynchronous tasks sent to this queue execute as they would in an `actor` type, allowing for re-entrancy and non-FIFO behavior when an individual task suspends.
+///
+/// An `ActorQueue` is used to ensure tasks sent from a nonisolated context to a single `actor`'s isolated context begin execution in order.
+/// Here is an example of how an `ActorQueue` should be utilized within an `actor`:
+/// ```
```
bachand, line 29:
```suggestion
/// ```swift
```

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -0,0 +1,144 @@
...
+                continuation.resume(returning: await task())
+            }
+        }
+    }
+
+    /// Schedules an asynchronous throwing task and returns after the task is complete.
+    /// The scheduled task will not execute until all prior tasks have completed or suspended.
+    /// - Parameter task: The task to enqueue.
+    /// - Returns: The value returned from the enqueued task.
+    public func await<T>(_ task: @escaping @Sendable () async throws -> T) async throws -> T {
```
bachand, line 106: If this function is accepts a throwing task should the above one accept a throwing task too?

Tests/AsyncQueueTests/ActorQueueTests.swift
```diff
@@ -0,0 +1,178 @@
...
+
+    func test_async_startsExecutionOfNextTaskAfterSuspension() async {
+        let semaphore = Semaphore()
+        systemUnderTest.async {
+            await semaphore.wait()
+        }
+        systemUnderTest.async {
+            // Signal the semaphore from the actor queue.
+            // If the actor queue were FIFO, this test would hang since this code would never execute:
+            // we'd still be waiting for the prior `wait()` tasks to finish.
```
bachand, line 57: 👍

bachand (approved):
Nice work @dfed !

Sources/AsyncQueue/ActorQueue.swift
dfed: Good catch! 071952

Sources/AsyncQueue/ActorQueue.swift
dfed: Ideally we'd use `rethrows` here so we only need to declare one method, however I couldn't get `rethrows` to work here: there's no way today (that I know of) to signal to the compiler that we're only going to `throw` within `withUnsafeThrowingContinuation` if and only if the `task` `throws`. So I created one method that can `throw` and one that can't.

README.md
```diff
@@ -6,26 +6,89 @@
...
 
-## Usage
+## Task Ordering and Swift Concurrency
 
-### Basic Initialization
+Tasks sent from a synchronous context to an asynchronous context in Swift Concurrency are inherently unordered. Consider the following test:
 
-```swift
-let asyncQueue = AsyncQueue()
 ```
```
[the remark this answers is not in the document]
dfed: Thanks for the prompt to do so!

PR #5, opened 2022-12-10
Add tests for Semaphore type

dfed:
This PR adds tests for the `Semaphore` type used in unit tests.

These tests aren't strictly necessary since the `Semaphore` type only lives under the `Tests` directory. But I found writing these tests instructive, so I figured they'd be worthwhile to land in the repo as prior art.

### DIFF
--- a/Tests/AsyncQueueTests/SemaphoreTests.swift
+++ b/Tests/AsyncQueueTests/SemaphoreTests.swift
@@ -33,10 +33,10 @@ final class SemaphoreTests: XCTestCase {
     }
 
     override func tearDown() async throws {
-        try await super.tearDown()
-
         let isWaiting = await systemUnderTest.isWaiting
         XCTAssertFalse(isWaiting)
+
+        try await super.tearDown()
     }
 
     // MARK: Behavior Tests
--- a/Tests/AsyncQueueTests/SemaphoreTests.swift
+++ b/Tests/AsyncQueueTests/SemaphoreTests.swift
@@ -35,9 +35,8 @@ final class SemaphoreTests: XCTestCase {
     override func tearDown() async throws {
         try await super.tearDown()
 
-        while await systemUnderTest.isWaiting {
-            await systemUnderTest.signal()
-        }
+        let isWaiting = await systemUnderTest.isWaiting
+        XCTAssertFalse(isWaiting)
     }
 
     // MARK: Behavior Tests
--- a/Tests/AsyncQueueTests/SemaphoreTests.swift
+++ b/Tests/AsyncQueueTests/SemaphoreTests.swift
@@ -1,0 +1,137 @@
+// MIT License
+//
+// Copyright (c) 2022 Dan Federman
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import XCTest
+
+final class SemaphoreTests: XCTestCase {
+
+    // MARK: XCTestCase
+
+    override func setUp() async throws {
+        try await super.setUp()
+
+        systemUnderTest = Semaphore()
+    }
+
+    override func tearDown() async throws {
+        try await super.tearDown()
+
+        while await systemUnderTest.isWaiting {
+            await systemUnderTest.signal()
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


dfed:
Given the CI failure, pulling this back into Draft

codecov.yml
```diff
@@ -12,7 +12,3 @@ coverage:
       default:
         threshold: 0.25%
     patch: off
-
-ignore:
-  # This package is not shipped.
-  - "Tests"
```
dfed, line 18: It's a little silly to require 100% coverage in unit tests but given the complexity of this library I see some benefit.

We can always re-add this `ignore` if maintaining this level of coverage proves problematic. But I like striving for full coverage in a library like this.

codecov.yml
```diff
@@ -10,9 +10,5 @@ coverage:
   status:
     project:
       default:
-        threshold: 0.25%
+        target: 100%
```
dfed, line 13: Similarly, it's a little silly to set a target of 100% coverage. But again, this library is _complex_, and (thus far) can be written without any runtime assertions that would prove difficult to test.

Given how fundamental this repo will be for consumers, let's start out by holding ourselves to a high standard.

dfed:
Figured it out! 3cdb8b0ba2c7f9595d75b28ea3f03729d09c0234 solved the issue, and then I went overboard with unit test coverage requirements in the following commits. Marking as ready for review!

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -0,0 +1,144 @@
...
+
+// MARK: - Semaphore Extension
+
+private extension Semaphore {
+    /// Enqueues an asynchronous task. This method suspends the caller until the asynchronous task has begun, ensuring ordered execution of enqueued tasks.
+    /// - Parameter task: A unit of work that returns work to execute after the task completes and the count is incremented.
+    func enqueueAndCount(using counter: UnsafeCounter, _ task: @escaping @Sendable (isolated Semaphore) async -> ((isolated Semaphore) -> Void)?) async {
+        // Await the start of the soon-to-be-enqueued `Task` with a continuation.
+        await withCheckedContinuation { continuation in
+            // Re-enter the actor context but don't wait for the result.
```
bachand, line 124: To make sure I follow... the actor context is the semaphore's context? And the reason that this task is on that context is because this method is isolated to the semaphore by the nature of it being an extension on an actor?

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -0,0 +1,144 @@
...
+    func execute(_ task: @Sendable (isolated Semaphore) async throws -> Void) async rethrows {
+        try await task(self)
+    }
+}
+
+// MARK: - UnsafeCounter
+
+private final class UnsafeCounter: @unchecked Sendable {
+    var countedTasksCompleted = 0
+}
```
bachand, line 144: It's OK that this isn't thread-safe since it's only accessed on the actor's context. Is that right?

bachand (approved):
I'm glad we're focused on solid test coverage 💪

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -0,0 +1,144 @@
...
+        try await super.setUp()
+
+        systemUnderTest = Semaphore()
+    }
+
+    override func tearDown() async throws {
+        try await super.tearDown()
+
+        let isWaiting = await systemUnderTest.isWaiting
+        XCTAssertFalse(isWaiting)
```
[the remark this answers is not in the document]
dfed: I continue to strive to call `super` unless documentation explicitly states that I _should not_ call `super` – I have previously found that not calling `super` when there is an `override` keyword makes me blind to the spots where I _should_ be calling Super. But yes, I should call this one at the end!

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -0,0 +1,144 @@
...
+         This test is tricky to pull off!
+         Our requirements:
+         1. We need to call `wait()` before `signal()`
+         2. We need to ensure that the `wait()` call suspends _before_ we call `signal()`
+         3. We can't `await` the `wait()` call before calling `signal()` since that would effectively deadlock the test.
+         4. We must utilize a single actor's isolated context to avoid accidental interleaving when suspending to communicate across actor contexts.
+
+         In order to ensure that we are executing the `wait()` calls before we call `signal()` _without awaiting a `wait()` call_,
+         we utilize an Actor (which has inherently ordered execution) to enqueue ordered `Task`s. We make calls to this actor
+         `await` the beginning of the `Task` to ensure that each `Task` has begun before resuming the test's execution.
```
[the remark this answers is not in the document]
dfed: welp. Thanks for catching! I cleaned it up a bit in 4a2848e, but tbh I'm not entirely sure what I was originally trying to say there.

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -0,0 +1,144 @@
...
+
+    // MARK: Behavior Tests
+
+    func test_wait_suspendsUntilEqualNumberOfSignalCalls() async {
+        /*
+         This test is tricky to pull off!
+         Our requirements:
+         1. We need to call `wait()` before `signal()`
+         2. We need to ensure that the `wait()` call suspends _before_ we call `signal()`
+         3. We can't `await` the `wait()` call before calling `signal()` since that would effectively deadlock the test.
```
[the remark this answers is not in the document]
dfed: yeah we can't `await` the `wait()` on the test's running queue. Updated in 519e1ed

Tests/AsyncQueueTests/SemaphoreTests.swift
dfed: Correct!

Tests/AsyncQueueTests/SemaphoreTests.swift
bachand: Curious to get your thoughts on this ^ when you have a moment @dfed . Tagging you here in case you missed this one.

dfed: I did miss this!

> the actor context is the semaphore's context

yes!

> the reason that this task is on that context is because this method is isolated to the semaphore by the nature of it being an extension on an actor

yes!

Thoughts on changing "the actor context" to "the semaphore context"?

bachand: > Thoughts on changing "the actor context" to "the semaphore context"?

Yeah I think that would increase clarity 👍

dfed: 4f38bfe

PR #6, opened 2022-12-11
Enable executing FIFO tasks on a desired actor's execution context

dfed:
This PR adds API to the `FIFOQueue` that enables `task`s sent to the queue to execute within a particular `actor`'s context.

The benefits of this API are:
1. Multiple interactions with a single `actor`-conforming type can be executed without requiring multiple suspensions.
2. Only interacting with `async`-decorated properties and methods on a given `actor` requires the `await` keyword, enabling the casual reader to distinguish which tasks will execute synchronously and which may suspend.

### DIFF
--- a/Sources/AsyncQueue/FIFOQueue.swift
+++ b/Sources/AsyncQueue/FIFOQueue.swift
@@ -59,7 +59,7 @@ public final class FIFOQueue: Sendable {
     }
 
     /// Schedules an asynchronous task for execution and immediately returns.
-    /// The scheduled task will not execute until all prior tasks have completed or suspended.
+    /// The scheduled task will not execute until all prior tasks have completed.
     /// - Parameters:
     ///   - isolatedActor: The actor within which the task is isolated.
     ///   - task: The task to enqueue.
@@ -80,7 +80,7 @@ public final class FIFOQueue: Sendable {
     }
 
     /// Schedules an asynchronous task and returns after the task is complete.
-    /// The scheduled task will not execute until all prior tasks have completed or suspended.
+    /// The scheduled task will not execute until all prior tasks have completed.
     /// - Parameters:
     ///   - isolatedActor: The actor within which the task is isolated.
     ///   - task: The task to enqueue.
@@ -110,7 +110,7 @@ public final class FIFOQueue: Sendable {
     }
 
     /// Schedules an asynchronous throwing task and returns after the task is complete.
-    /// The scheduled task will not execute until all prior tasks have completed or suspended.
+    /// The scheduled task will not execute until all prior tasks have completed.
     /// - Parameters:
     ///   - isolatedActor: The actor within which the task is isolated.
     ///   - task: The task to enqueue.
--- a/Sources/AsyncQueue/FIFOQueue.swift
+++ b/Sources/AsyncQueue/FIFOQueue.swift
@@ -58,6 +58,15 @@ public final class FIFOQueue: Sendable {
         taskStreamContinuation.yield(task)
     }
 
+    /// Schedules an asynchronous task for execution and immediately returns.
+    /// The scheduled task will not execute until all prior tasks have completed or suspended.
+    /// - Parameters:
+    ///   - isolatedActor: The actor within which the task is isolated.
+    ///   - task: The task to enqueue.
+    public func async<ActorType: Actor>(on isolatedActor: ActorType, _ task: @escaping @Sendable (isolated ActorType) async -> Void) {
+        taskStreamContinuation.yield { await task(isolatedActor) }
+    }
+
     /// Schedules an asynchronous task and returns after the task is complete.
     /// The scheduled task will not execute until all prior tasks have completed.
     /// - Parameter task: The task to enqueue.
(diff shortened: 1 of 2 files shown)


Tests/AsyncQueueTests/FIFOQueueTests.swift
```diff
@@ -65,7 +91,26 @@ final class FIFOQueueTests: XCTestCase {
         await systemUnderTest.await { /* Drain the queue */ }
     }
 
-    func test_async_isNotReentrant() async {
```
dfed, line 68: this diff is slightly confusingly presented: I renamed this test to be more clear, and added a new test above this one.

Tests/AsyncQueueTests/FIFOQueueTests.swift
```diff
@@ -65,7 +91,54 @@ final class FIFOQueueTests: XCTestCase {
...
+            }
+            counter.incrementAndExpectCount(equals: 1)
+            systemUnderTest.async(on: counter) { counter in
+                counter.incrementAndExpectCount(equals: 3)
+            }
+        }
+        await systemUnderTest.await { /* Drain the queue */ }
+    }
+
+    func test_await_async_areNotReentrant() async {
```
dfed, line 141: this test was previously called `test_async_isNotReentrant`, but technically it was testing that `await` and `async` aren't reentrant rather than just testing that `async` is not reentrant.

Tests/AsyncQueueTests/FIFOQueueTests.swift
```diff
@@ -46,6 +46,32 @@ final class FIFOQueueTests: XCTestCase {
         await systemUnderTest.await { /* Drain the queue */ }
     }
 
+    func test_asyncOn_sendsEventsInOrder() async {
```
dfed, line 49: Most of these new tests mirror the tests directly above them that tested the API without the `on isolatedActor` parameter.

Tests/AsyncQueueTests/FIFOQueueTests.swift
```diff
@@ -65,7 +91,54 @@ final class FIFOQueueTests: XCTestCase {
...
+                let isWaiting = semaphore.isWaiting
+                // This test will fail occasionally if we aren't executing atomically.
+                // You can prove this to yourself by replacing `systemUnderTest.async` above with `Task`.
+                XCTAssertFalse(isWaiting)
+                // Signal the semaphore before or after we wait – let the scheduler decide.
+                Task {
+                    semaphore.signal()
+                }
+                // Wait for the concurrent task to complete.
+                await semaphore.wait()
```
bachand, line 107: Here we still need an `await` since `wait()` may suspend. Is that right?

Tests/AsyncQueueTests/FIFOQueueTests.swift
```diff
@@ -152,12 +320,38 @@ final class FIFOQueueTests: XCTestCase {
...
+            }
+
+            guard iteration % 25 == 0 else {
+                // Keep sending async events to the queue.
+                continue
+            }
+
+            await systemUnderTest.await(on: counter) { counter in
+                let count = counter.count
+                XCTAssertEqual(count, iteration)
```
bachand, line 337: 👍

bachand (approved):
Nice work @dfed !

Sources/AsyncQueue/FIFOQueue.swift
```diff
@@ -58,6 +58,15 @@ public final class FIFOQueue: Sendable {
         taskStreamContinuation.yield(task)
     }
 
+    /// Schedules an asynchronous task for execution and immediately returns.
+    /// The scheduled task will not execute until all prior tasks have completed or suspended.
```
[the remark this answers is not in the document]
dfed: oops. Copy paste! 56549e2

Sources/AsyncQueue/FIFOQueue.swift
```diff
@@ -70,6 +79,20 @@ public final class FIFOQueue: Sendable {
         }
     }
 
+    /// Schedules an asynchronous task and returns after the task is complete.
+    /// The scheduled task will not execute until all prior tasks have completed or suspended.
```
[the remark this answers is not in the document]
dfed: 56549e2

Sources/AsyncQueue/FIFOQueue.swift
```diff
@@ -86,6 +109,24 @@ public final class FIFOQueue: Sendable {
         }
     }
 
+    /// Schedules an asynchronous throwing task and returns after the task is complete.
+    /// The scheduled task will not execute until all prior tasks have completed or suspended.
```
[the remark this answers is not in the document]
dfed: 56549e2

Tests/AsyncQueueTests/FIFOQueueTests.swift
dfed: Correct! `wait` is an `async` (i.e. possibly suspending) method on the semaphore, whereas `signal` is not. So we need to `await` the `async` method.

PR #7, opened 2022-12-11
Make ActorQueue's tasks isolated to an actor's context

dfed:
This PR is built on #3

This PR is similar to #6 but applied to the `ActorQueue`. However, rather than adding API, this PR replaces existing `async` and `await` API with methods where the `task`s sent to the queue to execute on the designated `actor`'s context.

The goals of this PR are:
1. Enable multiple calls to the sam actor within a single `task` without requiring multiple suspension points (`await` calls) within the `task
2. Ensure that `await` calls within a single `task` actually result in suspension

These goals are important because an `ActorQueue`'s `task` allows other tasks to begin execution as soon as the `task` suspends (via an `await`). Since `ActorQueue`s should only be utilized with a single `actor` instance, it made sense to replace the existing API rather than add new API – there's no good reason (that I can come up with) to use the the replaced APIs now that these new APIs exist.

All that said, I don't love this API change. `ActorQueue`s are meant to be used as private implementation details of an `actor`, and as such the `isolatedActor` that will be sent to the `ActorQueue`'s API is almost always going to be `self`. The big drawback of this change is that Swift doesn't allow parameters to a closure to be named `self`, so `ActorQueue` calls will likely end up looking like the following:
```
queue.async(on: self) { myself in
    myself.doSomething()
}
```

If someone were to try to ignore the parameter in the closure, they'd end up having to write an `await`:
```
queue.async(on: self) { _ in
    // This `task` is isolated to the parameter, which Swift doesn't know is the same as `self`.
    // The result is that Swift requires us to add an `await` here even though we're executing within `self`'s context.
    await self.doSomething()
}
```

I really like the goals I'm trying to achieve, but I don't love how the semantics play out. Curious to get your thoughts!

### DIFF
--- a/Tests/AsyncQueueTests/ActorQueueTests.swift
+++ b/Tests/AsyncQueueTests/ActorQueueTests.swift
@@ -39,34 +39,34 @@ final class ActorQueueTests: XCTestCase {
     func test_async_sendsEventsInOrder() async {
         let counter = Counter()
         for iteration in 1...1_000 {
-            systemUnderTest.async {
-                await counter.incrementAndExpectCount(equals: iteration)
+            systemUnderTest.async(on: counter) { counter in
+                counter.incrementAndExpectCount(equals: iteration)
             }
         }
-        await systemUnderTest.await { /* Drain the queue */ }
+        await systemUnderTest.await(on: counter) { _ in /* Drain the queue */ }
     }
 
     func test_async_startsExecutionOfNextTaskAfterSuspension() async {
         let semaphore = Semaphore()
-        systemUnderTest.async {
+        systemUnderTest.async(on: semaphore) { semaphore in
             await semaphore.wait()
         }
-        systemUnderTest.async {
+        systemUnderTest.async(on: semaphore) { semaphore in
             // Signal the semaphore from the actor queue.
             // If the actor queue were FIFO, this test would hang since this code would never execute:
             // we'd still be waiting for the prior `wait()` tasks to finish.
-            await semaphore.signal()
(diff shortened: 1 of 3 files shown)


Tests/AsyncQueueTests/ActorQueueTests.swift
```diff
@@ -39,47 +39,46 @@ final class ActorQueueTests: XCTestCase {
...
         let counter = Counter()
         for iteration in 1...1_000 {
-            systemUnderTest.async {
-                await counter.incrementAndExpectCount(equals: iteration)
+            systemUnderTest.async(on: counter) { counter in
+                counter.incrementAndExpectCount(equals: iteration)
             }
         }
-        await systemUnderTest.await { /* Drain the queue */ }
+        await systemUnderTest.await(on: counter) { _ in /* Drain the queue */ }
```
dfed, line 46: I'm tempted to add a new `internal` method called `func allEnqueuedTasksSuspendedOrCompleted() async` to make this test call-site read more clearly since we have a few of these.

dfed:
Removing reviewers and bringing this PR back to drafts. Had some ideas I want to play with – I may be able to simplify call-sites make the async for-loop run on the target actor's context.

dfed:
Ideally the ActorQueue would run entirely on its target actor's execution context, but that's not possible today given [how `Task`s and `isolated` parameters interact today](https://forums.swift.org/t/should-task-inherit-the-execution-context-of-an-isolated-parameter/61999).

I'm going to close this out for now, though I'll want to explore this path more in the future.

PR #8, opened 2023-01-12
Add consistent examples to README

dfed:
This PR gets each section of the README re-visiting the same ordering test to make explicit how these types solve the problem statement in the `Task Ordering and Swift Concurrency` section of the README.

There's still room to improve the explanation of these types, but we're making progress here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,13 +14,21 @@ Tasks sent from a synchronous context to an asynchronous context in Swift Concur
 
 ```swift
 @MainActor
-func test_mainActor_taskOrdering() async {
-    var counter = 0
+func testMainActorTaskOrdering() async {
+    actor Counter {
+        func increment() -> Int {
+            count += 1
+            return count
+        }
+        var count = 0
+    }
+
+    let counter = Counter()
     var tasks = [Task<Void, Never>]()
     for iteration in 1...100 {
         tasks.append(Task {
-            counter += 1
-            XCTAssertEqual(counter, iteration) // often fails
+            let incrementedCount = await counter.increment()
+            XCTAssertEqual(incrementedCount, iteration) // often fails
         })
     }
     for task in tasks {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,13 +14,21 @@ Tasks sent from a synchronous context to an asynchronous context in Swift Concur
 
 ```swift
 @MainActor
-func test_mainActor_taskOrdering() async {
-    var counter = 0
+func testMainActorTaskOrdering() async {
+    actor Counter {
```
dfed, line 18: We don't _need_ a `Counter` type in this example but the below examples do need it. For the sake of consistency with the below additions I made this example test a dash more complicated.

README.md
```diff
@@ -51,14 +59,35 @@ queue.async {
     This task begins execution once the above one-second sleep completes.
     */
 }
-Task {
```
dfed, line 54: This `Task` was meant to show that we were entering an async context, but I'm realizing now that the inclusion of the `Task` is more confusing than not. Nothing within the `Task` closure changed – I just removed the `Task` enclosure. I did the same thing further down as well.

README.md
```diff
@@ -80,14 +109,35 @@ queue.async {
...
+    let queue = ActorQueue()
+    for iteration in 1...100 {
+        queue.async {
+            let incrementedCount = await counter.increment()
+            XCTAssertEqual(incrementedCount, iteration) // always succeeds
+        }
     }
+    await queue.await { }
 }
 ```
```
dfed, line 142: Ideally we'd explain better how the `ActorQueue` and `FIFOQueue` differ in the above example test, but I'm wary of trying to communicate too many ideas at once.

README.md
bachand: This makes sense to me. The reader can infer from the `await` below that we're in an asynchronous context.

README.md
```diff
@@ -51,14 +59,35 @@ queue.async {
...
+
+    let counter = Counter()
+    let queue = FIFOQueue()
+    for iteration in 1...100 {
+        queue.async {
+            let incrementedCount = await counter.increment()
+            XCTAssertEqual(incrementedCount, iteration) // always succeeds
+        }
     }
+    await queue.await { }
```
bachand, line 90: Maybe clarify with a comment that we're ensuring all tasks complete before we end the test?

README.md
```diff
@@ -80,14 +109,35 @@ queue.async {
...
+
+    let counter = Counter()
+    let queue = ActorQueue()
+    for iteration in 1...100 {
+        queue.async {
+            let incrementedCount = await counter.increment()
+            XCTAssertEqual(incrementedCount, iteration) // always succeeds
+        }
     }
+    await queue.await { }
```
bachand, line 140: Maybe we also clarify the purpose of this line.

bachand (approved):

README.md
dfed: I'll fix this in #9 which builds on top of this change.

README.md
dfed: I'll fix this in #9 which builds on top of this change.

README.md
dfed: c4d455f

PR #9, opened 2023-01-16
Make ActorQueue tasks execute on the target actor's execution context

dfed:
This PR updates the `ActorQueue` API to make `task`s sent to the queue to execute within the target `actor`'s context. This API change is a breaking one, and so we'll be bumping our version to `0.2.0` as part of this change.

Similarly to the benefits discussed in #6, the benefits of running tasks on the target `actor`'s context are:
1. Multiple interactions with a single `actor`-conforming type can be executed without requiring multiple suspensions.
2. Only interacting with `async`-decorated properties and methods on a given `actor` requires the `await` keyword, enabling the casual reader to distinguish which tasks will execute synchronously and which may suspend.

These benefits are particularly important on an `ActorQueue`, where suspension can lead to interleaving of enqueued work.

The downside of this API is that avoiding a retain cycle between an actor queue and the target actor requires the use of an unowned reference and some force unwrapping. See comments below.

Note that this PR is built on top of #8 to avoid conflicts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ func testMainActorTaskOrdering() async {
             await counter.incrementAndAssertCountEquals(iteration)
         })
     }
+    // Wait for all enqueued tasks to finish.
     for task in tasks {
         _ = await task.value
     }
@@ -77,6 +78,7 @@ func testFIFOQueueOrdering() async {
     for iteration in 1...100 {
         counter.incrementAndAssertCountEquals(iteration)
     }
+    // Wait for all enqueued tasks to finish.
     await counter.flushQueue()
 }
 ```
@@ -118,6 +120,7 @@ func testActorQueueOrdering() async {
     for iteration in 1...100 {
         counter.incrementAndAssertCountEquals(iteration)
     }
+    // Wait for all enqueued tasks to finish.
     await counter.flushQueue()
 }
 ```
--- a/README.md
+++ b/README.md
@@ -91,53 +91,63 @@ func testFIFOQueueOrdering() async {
 }
 ```
 
-### Sending ordered asynchronous tasks to Actors
+### Sending ordered asynchronous tasks to Actors from a nonisolated context
 
-Use an `ActorQueue` to send ordered asynchronous tasks from a nonisolated context to an `actor` instance. Tasks sent to one of these queues are guaranteed to begin executing in the order in which they are enqueued. Ordering of execution is guaranteed up until the first [suspension point](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html#ID639) within the called `actor` code.
+Use an `ActorQueue` to send ordered asynchronous tasks to an `actor`'s isolated context from a nonisolated and synchronous context. Tasks sent to an actor queue are guaranteed to begin executing in the order in which they are enqueued. Ordering of execution is guaranteed up until the first [suspension point](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html#ID639) within the enqueued task.
 
 ```swift
 let queue = ActorQueue()
-queue.async {
+queue.setTargetContext(to: actor)
+queue.async { actor in
     /*
     `async` context that executes after all other enqueued work has begun executing.
     Work enqueued after this task will wait for this task to complete or suspend.
     */
-    try? await Task.sleep(nanoseconds: 1_000_000)
+    await actor.longRunningTask()
 }
-queue.async {
+queue.async { actor in
     /*
-    This task begins execution once the above task suspends due to the one-second sleep.
+    This task begins execution once the above task suspends due to the long-running task.
     */
 }
-await queue.await {
+await queue.await { actor in
     /*
     `async` context that can return a value or throw an error.
     Executes after all other enqueued work has begun executing.
     Work enqueued after this task will wait for this task to complete or suspend.
     */
 }
-```
 
 With an `ActorQueue` you can easily begin execution of asynchronous tasks from a nonisolated context in order:
 ```swift
 func testActorQueueOrdering() async {
     actor Counter {
-        func increment() -> Int {
-            count += 1
-            return count
+        init() {
+            queue.setTargetContext(to: self)
+        }
+
+        nonisolated
+        func increment(expectedCount: Int) {
+            queue.async { myself in
+                myself.count += 1
+                XCTAssertEqual(expectedCount, myself.count) // always succeeds
+            }
         }
         var count = 0
+
+        nonisolated
+        func flushQueue() async {
+            await queue.await { _ in }
+        }
+
+        private let queue = ActorQueue<Counter>()
     }
 
     let counter = Counter()
-    let queue = ActorQueue()
     for iteration in 1...100 {
-        queue.async {
-            let incrementedCount = await counter.increment()
-            XCTAssertEqual(incrementedCount, iteration) // always succeeds
-        }
+        counter.increment(expectedCount: iteration)
     }
-    await queue.await { }
+    await counter.flushQueue()
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -89,13 +89,16 @@ FIFO execution has a key downside: the queue must wait for all previously enqueu
 
 Use an `ActorQueue` to send ordered asynchronous tasks to an `actor`’s isolated context from nonisolated or synchronous contexts. Tasks sent to an actor queue are guaranteed to begin executing in the order in which they are enqueued. However, unlike a `FIFOQueue`, execution order is guaranteed only until the first [suspension point](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html#ID639) within the enqueued task. An `ActorQueue` executes tasks within the its adopted actor’s isolated context, resulting in `ActorQueue` task execution having the same properties as `actor` code execution: code between suspension points is executed atomically, and tasks sent to a single `ActorQueue` can await results from the queue without deadlocking.
 
-An instance of an `ActorQueue` is designed to be utilized by a single `actor` instance: tasks sent to an `ActorQueue` utilize the isolated context of the queue‘s adopted `actor` to serialize tasks. As such, the lifecycle of any `ActorQueue` should not exceed the lifecycle of its `actor`. It is strongly recommended that an `ActorQueue` be a `let` constant on the adopted `actor`. Additionally, an `actor` utilizing an `ActorQueue` should set the adopted execution context of the queue to `self` within the `actor`’s `init`. Failing to set an adopted execution context prior to enqueuing work on an `ActorQueue` will result in a crash.
+An instance of an `ActorQueue` is designed to be utilized by a single `actor` instance: tasks sent to an `ActorQueue` utilize the isolated context of the queue‘s adopted `actor` to serialize tasks. As such, there are a few requirements that must be met when dealing with an `ActorQueue`:
+1. The lifecycle of any `ActorQueue` should not exceed the lifecycle of its `actor`. It is strongly recommended that an `ActorQueue` be a `let` constant on the adopted `actor`. Enqueuing a task to an `ActorQueue` isntance after its adopted `actor` has been deallocated will result in a crash.
+2. An `actor` utilizing an `ActorQueue` should set the adopted execution context of the queue to `self` within the `actor`’s `init`. Failing to set an adopted execution context prior to enqueuing work on an `ActorQueue` will result in a crash.
 
 An `ActorQueue` can easily enqueue tasks that execute on an actor’s isolated context from a nonisolated context in order:
 ```swift
 func testActorQueueOrdering() async {
     actor Counter {
         init() {
+            // Adopting the execution context in `init` satisfies requirement #2 above.
             queue.adoptExecutionContext(of: self)
         }
 
@@ -113,6 +116,7 @@ func testActorQueueOrdering() async {
         }
 
         private var count = 0
+        // Making the queue a private let constant satisfies requirement #1 above.
         private let queue = ActorQueue<Counter>()
     }
 
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ While [actors](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html#
 
 ### Executing asynchronous tasks in FIFO order
 
-Use a `FIFOQueue` to execute asynchronous tasks enqueued from a nonisolated context in FIFO order. Tasks sent to one of these queues are guaranteed to begin _and end_ executing in the order in which they are enqueued. A `FIFOQueue` executes tasks in a similar manner to a `DispatchQueue`: enqueued tasks executes atomically, and a `FIFOQueue` task will deadlock the queue if it awaits results from the queue on which it is executing.
+Use a `FIFOQueue` to execute asynchronous tasks enqueued from a nonisolated context in FIFO order. Tasks sent to one of these queues are guaranteed to begin _and end_ executing in the order in which they are enqueued. A `FIFOQueue` executes tasks in a similar manner to a `DispatchQueue`: enqueued tasks executes atomically, and the program will deadlock if a task executing on a `FIFOQueue` awaits results from the queue on which it is executing.
 
 A `FIFOQueue` can easily execute asynchronous tasks from a nonisolated context in FIFO order:
 ```swift
--- a/README.md
+++ b/README.md
@@ -16,19 +16,20 @@ Tasks sent from a synchronous context to an asynchronous context in Swift Concur
 @MainActor
 func testMainActorTaskOrdering() async {
     actor Counter {
-        func increment() -> Int {
+        func incrementAndAssertCountEquals(_ expectedCount: Int) {
             count += 1
-            return count
+            let incrementedCount = count
+            XCTAssertEqual(incrementedCount, expectedCount) // often fails
         }
-        var count = 0
+
+        private var count = 0
     }
 
     let counter = Counter()
     var tasks = [Task<Void, Never>]()
     for iteration in 1...100 {
         tasks.append(Task {
-            let incrementedCount = await counter.increment()
-            XCTAssertEqual(incrementedCount, iteration) // often fails
+            await counter.incrementAndAssertCountEquals(iteration)
         })
     }
     for task in tasks {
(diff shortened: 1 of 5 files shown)


Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -23,53 +23,52 @@
...
 /// Asynchronous tasks sent to this queue execute as they would in an `actor` type, allowing for re-entrancy and non-FIFO behavior when an individual task suspends.
+/// Tasks execute within the target actor's isolated context
 ///
 /// An `ActorQueue` is used to ensure tasks sent from a nonisolated context to a single `actor`'s isolated context begin execution in order.
 /// Here is an example of how an `ActorQueue` should be utilized within an `actor`:
 /// ```swift
 /// public actor LogStore {
 ///
+///     public init() {
+///         queue.setTargetContext(to: self)
```
dfed, line 34: I don't love the `setTargetContext(to:)` method signature. I welcome better suggestions!

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -80,65 +78,78 @@ public final class ActorQueue {
...
+        target = actor
+    }
+
     /// Schedules an asynchronous task for execution and immediately returns.
     /// The scheduled task will not execute until all prior tasks have completed or suspended.
     /// - Parameter task: The task to enqueue.
-    public func async(_ task: @escaping @Sendable () async -> Void) {
-        taskStreamContinuation.yield(task)
+    public func async(_ task: @escaping @Sendable (isolated ActorType) async -> Void) {
+        taskStreamContinuation.yield(ActorTask(target: target!, task: task))
```
dfed, line 92: Using force unwraps is fine here given the strict instructions for how an `ActorQueue` should be utilized:
* An `ActorQueue`'s lifecycle should not exceed the lifecycle of its `target`
* An `ActorQueue`'s `target` should be set in the `target` actor's `init`

Therefore, the `target` will always be available if the API is properly utilized.

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -80,65 +78,78 @@ public final class ActorQueue {
...
 
-    private let taskStreamContinuation: AsyncStream<@Sendable () async -> Void>.Continuation
+    private let taskStreamContinuation: AsyncStream<ActorTask>.Continuation
 
-    // MARK: - ActorExecutor
+    /// The target actor on whose isolated context our tasks run.
+    /// It is safe to use `unowned` here because it is API misuse to interact with an `ActorQueue` from an instance other than the `target`.
+    private unowned var target: ActorType?
 
-    private actor ActorExecutor {
```
dfed, line 124: Now that we're explicitly targeting a specific actor, we no longer need another `actor` type! The code previously on `ActorExecutor` moved down into an extension on `Actor` below.

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -80,65 +78,78 @@ public final class ActorQueue {
...
-        func suspendUntilStarted(_ task: @escaping @Sendable () async -> Void) async {
-            // Suspend the calling code until our enqueued task starts.
-            await withUnsafeContinuation { continuation in
-                // Utilize the serial (but not FIFO) Actor context to execute the task without requiring the calling method to wait for the task to complete.
-                Task {
-                    // Force this task to execute within the ActorExecutor's context by accessing an ivar on the instance.
-                    // This works around a bug when compiling with Xcode 14.1: https://github.com/apple/swift/issues/62503
-                    _ = void
```
dfed, line 132: We no longer need this since we're sending `self` as a parameter of the `task`

README.md
```diff
@@ -43,101 +43,81 @@ While [actors](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html#
 
 ### Executing asynchronous tasks in FIFO order
 
-Use a `FIFOQueue` to execute asynchronous tasks enqueued from a nonisolated context in FIFO order. Tasks sent to one of these queues are guaranteed to begin _and end_ executing in the order in which they are enqueued.
+Use a `FIFOQueue` to execute asynchronous tasks enqueued from a nonisolated context in FIFO order. Tasks sent to one of these queues are guaranteed to begin _and end_ executing in the order in which they are enqueued. A `FIFOQueue` executes tasks in a similar manner to a `DispatchQueue`: enqueued tasks executes atomically, and a `FIFOQueue` task will deadlock the queue if it awaits results from the queue on which it is executing.
 
-```swift
```
dfed, line 48: This code block didn't add as much as I wanted it to, and with the new API rev on `ActorQueue` it was difficult to showcase the `ActorQueue` API in a similar manner without breaking the `ActorQueue`'s guidelines (i.e. we couldn't show a concise, free-form actor queue without adopting the queue execution context outside of an actor's `init`)

README.md
```diff
@@ -43,101 +45,83 @@ While [actors](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html#
...
+
+An instance of an `ActorQueue` is designed to be utilized by a single `actor` instance: tasks sent to an `ActorQueue` utilize the isolated context of the queue‘s adopted `actor` to serialize tasks. As such, the lifecycle of any `ActorQueue` should not exceed the lifecycle of its `actor`. It is strongly recommended that an `ActorQueue` be a `let` constant on the adopted `actor`. Additionally, an `actor` utilizing an `ActorQueue` should set the adopted execution context of the queue to `self` within the `actor`’s `init`. Failing to set an adopted execution context prior to enqueuing work on an `ActorQueue` will result in a crash.
```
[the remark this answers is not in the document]
bachand: Or you could make it a numbered list an reference the numbered requirements in the code below.

README.md
```diff
@@ -43,101 +45,83 @@ While [actors](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html#
...
-            return count
+        init() {
+            queue.adoptExecutionContext(of: self)
         }
-        var count = 0
+
+        nonisolated
+        func incrementAndAssertCountEquals(_ expectedCount: Int) {
+            queue.async { myself in
+                myself.count += 1
```
bachand, line 105: No `await` — nice!

README.md
```diff
@@ -43,101 +45,83 @@ While [actors](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html#
...
-    let queue = ActorQueue()
     for iteration in 1...100 {
-        queue.async {
-            let incrementedCount = await counter.increment()
-            XCTAssertEqual(incrementedCount, iteration) // always succeeds
-        }
+        counter.incrementAndAssertCountEquals(iteration)
     }
-    await queue.await { }
+    // Wait all enqueued tasks to finish.
```
bachand, line 123:
```suggestion
    // Wait for all enqueued tasks to finish.
```

README.md
```diff
@@ -43,101 +45,83 @@ While [actors](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html#
...
-    `async` context that can return a value or throw an error.
-    Executes after all other enqueued work is completed.
-    Work enqueued after this task will wait for this task to complete.
-    */
-}
-```
+Use a `FIFOQueue` to execute asynchronous tasks enqueued from a nonisolated context in FIFO order. Tasks sent to one of these queues are guaranteed to begin _and end_ executing in the order in which they are enqueued. A `FIFOQueue` executes tasks in a similar manner to a `DispatchQueue`: enqueued tasks executes atomically, and a `FIFOQueue` task will deadlock the queue if it awaits results from the queue on which it is executing.
 
-With a `FIFOQueue` you can easily execute asynchronous tasks from a nonisolated context in FIFO order:
+A `FIFOQueue` can easily execute asynchronous tasks from a nonisolated context in FIFO order:
```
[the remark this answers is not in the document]
bachand: That being said I see that we mention "asynchronous tasks" in many places in the actor queue section so it may not be a big deal.

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -20,56 +20,54 @@
...
 ///
-///     private let queue = ActorQueue()
+///     private let queue = ActorQueue<LogStore>()
 ///     private var logs = [String]()
 /// }
 /// ```
 ///
-/// - Warning: Execution order is not guaranteed unless the enqueued tasks interact with a single `actor` instance.
-public final class ActorQueue {
+/// - Warning: The lifecycle of an `ActorQueue` should not exceed that of the adopted actor.
```
bachand, line 53: What happens if it does? We may want to upgrade the severity to "Important" or "Precondition".

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -20,56 +20,54 @@
...
+public final class ActorQueue<ActorType: Actor> {
 
     // MARK: Initialization
 
     /// Instantiates an actor queue.
-    /// - Parameter priority: The baseline priority of the tasks added to the asynchronous queue.
-    public init(priority: TaskPriority? = nil) {
```
bachand, line 60: Why do we still allow a `FIFOQueue` to be initialized with a `priority` but not an actor queue? Does it have something to do with how we're using the actor's context for task execution?

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -80,65 +78,88 @@ public final class ActorQueue {
...
 
     // MARK: Private
 
-    private let taskStreamContinuation: AsyncStream<@Sendable () async -> Void>.Continuation
+    private let taskStreamContinuation: AsyncStream<ActorTask>.Continuation
 
-    // MARK: - ActorExecutor
+    /// The executionContext actor on whose isolated context our tasks run.
+    /// It is safe to use `unowned` here because it is API misuse to interact with an `ActorQueue` from an instance other than the `executionContext`.
+    private unowned var executionContext: ActorType?
```
bachand, line 144: Hmm why not use `weak`? It's been awhile since I thought about `unowned` but I thought `unowned` was useful if you wanted a non-optional variable. Since this is already optional is there any downside to using `weak`?

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -80,65 +78,88 @@ public final class ActorQueue {
...
     // MARK: Public
 
+    /// Sets the actor context within which each `async` and `await`ed task will execute.
+    /// It is recommended that this method be called in the adopted actor’s `init` method.
+    /// **Must be called prior to enqueuing any work on the receiver.**
+    ///
+    /// - Parameter actor: The actor on which the queue's task will execute. This parameter is not retained by the receiver.
+    /// - Warning: Calling this method more than once will result in an assertion failure.
+    public func adoptExecutionContext(of actor: ActorType) {
+        assert(executionContext == nil) // Setting multiple executionContexts on the same queue is API abuse.
```
bachand, line 88: 👍

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -80,65 +78,88 @@ public final class ActorQueue {
...
+    }
+
     /// Schedules an asynchronous task for execution and immediately returns.
     /// The scheduled task will not execute until all prior tasks have completed or suspended.
-    /// - Parameter task: The task to enqueue.
-    public func async(_ task: @escaping @Sendable () async -> Void) {
-        taskStreamContinuation.yield(task)
+    /// - Parameter task: The task to enqueue. The task's parameter is a reference to the actor whose execution context has been adopted.
+    public func async(_ task: @escaping @Sendable (isolated ActorType) async -> Void) {
+        // Crashing here means that this queue is being sent tasks either before an execution context has been set, or
```
bachand, line 96: Nice helpful comment for a consumer of the library 👍

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -80,65 +78,88 @@ public final class ActorQueue {
...
     /// - Returns: The value returned from the enqueued task.
-    public func await<T>(_ task: @escaping @Sendable () async -> T) async -> T {
-        await withUnsafeContinuation { continuation in
-            taskStreamContinuation.yield {
-                continuation.resume(returning: await task())
-            }
+    public func await<T>(_ task: @escaping @Sendable (isolated ActorType) async -> T) async -> T {
+        // Crashing here means that this queue is being sent tasks either before an execution context has been set, or
+        // after the execution context has deallocated. An ActorQueue's execution context should be set in the adopted
+        // actor's `init` method, and the ActorQueue should not exceed the lifecycle of the adopted actor.
```
bachand, line 109: We could make a helper method or property that allow us to write this comment only once in this file.

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -80,65 +78,88 @@ public final class ActorQueue {
...
+        let executionContext = self.executionContext! // Capture/retain the executionContext before suspending.
```
bachand, line 110: If there is an asynchronous work pending and the actor is deallocated, we will keep the actor alive until the work is done 👌

Sources/AsyncQueue/FIFOQueue.swift
```diff
@@ -94,7 +94,7 @@ public final class FIFOQueue: Sendable {
     }
 
     /// Schedules an asynchronous throwing task and returns after the task is complete.
-    /// The scheduled task will not execute until all prior tasks have completed.
+    /// The scheduled task will not execute until all prior tasks – including suspended tasks – have completed.
```
bachand, line 97: I'm pretty sure we missed one https://github.com/dfed/swift-async-queue/blob/c4d455f4e9d8116e8e9479e35e3400f2f3f94734/Sources/AsyncQueue/FIFOQueue.swift#L113

Sources/AsyncQueue/ActorQueue.swift
bachand: This seems to be proven in `test_async_retainsAdoptedActorUntilQueueFlushes()`

Tests/AsyncQueueTests/ActorQueueTests.swift
```diff
@@ -31,54 +31,99 @@ final class ActorQueueTests: XCTestCase {
...
+    func test_adoptExecutionContext_doesNotRetainActor() {
+        let systemUnderTest = ActorQueue<Counter>()
+        var counter: Counter? = Counter()
+        weak var weakCounter = counter
+        systemUnderTest.adoptExecutionContext(of: counter!)
+        counter = nil
+        XCTAssertNil(weakCounter)
+    }
+
+    func test_async_retainsAdoptedActorUntilQueueFlushes() async {
```
bachand, line 50: This test doesn't flush the queue. Can we rename it to something about how the actor is retained while there is in-progress tasks?

Tests/AsyncQueueTests/ActorQueueTests.swift
```diff
@@ -31,54 +31,99 @@ final class ActorQueueTests: XCTestCase {
...
+        systemUnderTest.async { counter in
+            await semaphore.wait()
+        }
+
+        counter = nil
+        XCTAssertNotNil(weakCounter)
+        await semaphore.signal()
+    }
+
+    func test_async_taskParameterIsAdoptedActor() async {
```
bachand, line 66: 👍

Tests/AsyncQueueTests/ActorQueueTests.swift
```diff
@@ -31,54 +31,99 @@ final class ActorQueueTests: XCTestCase {
...
+
+        systemUnderTest.async { semaphore in
             await semaphore.wait()
         }
-        systemUnderTest.async {
+        systemUnderTest.async { semaphore in
             // Signal the semaphore from the actor queue.
             // If the actor queue were FIFO, this test would hang since this code would never execute:
             // we'd still be waiting for the prior `wait()` tasks to finish.
-            await semaphore.signal()
```
bachand, line 58: 🎉

Tests/AsyncQueueTests/ActorQueueTests.swift
```diff
@@ -135,28 +183,26 @@ final class ActorQueueTests: XCTestCase {
...
             guard iteration % 25 == 0 else {
                 // Keep sending async events to the queue.
                 continue
             }
 
-            await systemUnderTest.await {
-                let count = await counter.count
-                XCTAssertEqual(count, iteration)
+            await systemUnderTest.await { counter in
+                XCTAssertEqual(counter.count, iteration)
```
bachand, line 197: So much cleaner.

bachand (approved):
Love this @dfed . Great work iterating and coming to such an elegant solution 🏆

README.md
```diff
@@ -16,19 +16,20 @@ Tasks sent from a synchronous context to an asynchronous context in Swift Concur
 @MainActor
 func testMainActorTaskOrdering() async {
     actor Counter {
-        func increment() -> Int {
+        func incrementAndAssertCountEquals(_ expectedCount: Int) {
             count += 1
-            return count
+            let incrementedCount = count
+            XCTAssertEqual(incrementedCount, expectedCount) // often fails
```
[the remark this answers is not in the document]
dfed: > Curious why we need/want to change to this version of the code that no longer returns the new count.

In this example it doesn't matter much, but in later examples where we have code like the following it becomes necessary and I didn't want to change examples too much:
```
    for iteration in 1...100 {
        counter.incrementAndAssertCountEquals(iteration)
    }
```

If we used an `increment` that would return, we'd end up with:
```
    for iteration in 1...100 {
        let count = await counter.increment()
        XCTAssertEqual(count, iteration) // this test will succeed since we `await` the return value of `increment` each time
    }
```

README.md
```diff
@@ -16,21 +16,23 @@ Tasks sent from a synchronous context to an asynchronous context in Swift Concur
...
     let counter = Counter()
     var tasks = [Task<Void, Never>]()
     for iteration in 1...100 {
         tasks.append(Task {
-            let incrementedCount = await counter.increment()
-            XCTAssertEqual(incrementedCount, iteration) // often fails
+            await counter.incrementAndAssertCountEquals(iteration)
         })
     }
+    // Wait all enqueued tasks to finish.
```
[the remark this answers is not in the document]
dfed: 🤦‍♂️ yes. Fixed up in a4871ad

README.md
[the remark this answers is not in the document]
dfed: I don't think we want to qualify the tasks as async everywhere – would get reasonably verbose if we did that 😅

README.md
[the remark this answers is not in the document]
dfed: Love that idea! 2a63fc0

Sources/AsyncQueue/ActorQueue.swift
dfed: Good call. We will crash, so `precondition` seems like the right move ac49f33

Sources/AsyncQueue/ActorQueue.swift
dfed: > Does it have something to do with how we're using the actor's context for task execution?

Exactly this. We could create a high-priority for-loop but we're always going to be waiting on the actor's context anyways so it seems not terribly useful to have a `priority` argument here. We can always add it back.

Sources/AsyncQueue/ActorQueue.swift
dfed: 🎉 yeah I figure if I'm going to be using `unowned` and force unwraps I might as well guide the consumer when the API inevitably is abused.

Sources/AsyncQueue/ActorQueue.swift
dfed: Yes on both counts!

Sources/AsyncQueue/ActorQueue.swift
dfed: c92e62c

Sources/AsyncQueue/ActorQueue.swift
dfed: Entirely fair! 18a0183

Sources/AsyncQueue/FIFOQueue.swift
dfed: ahh I missed all the methods added in #7. cdfdfe6

Tests/AsyncQueueTests/ActorQueueTests.swift
dfed: Technically we are flushing the queue by waiting for the `async` task to complete, but that's not super intuitive! Updated in 1c8cab5

README.md
```diff
@@ -89,13 +89,16 @@ FIFO execution has a key downside: the queue must wait for all previously enqueu
...
 
-An instance of an `ActorQueue` is designed to be utilized by a single `actor` instance: tasks sent to an `ActorQueue` utilize the isolated context of the queue‘s adopted `actor` to serialize tasks. As such, the lifecycle of any `ActorQueue` should not exceed the lifecycle of its `actor`. It is strongly recommended that an `ActorQueue` be a `let` constant on the adopted `actor`. Additionally, an `actor` utilizing an `ActorQueue` should set the adopted execution context of the queue to `self` within the `actor`’s `init`. Failing to set an adopted execution context prior to enqueuing work on an `ActorQueue` will result in a crash.
+An instance of an `ActorQueue` is designed to be utilized by a single `actor` instance: tasks sent to an `ActorQueue` utilize the isolated context of the queue‘s adopted `actor` to serialize tasks. As such, there are a few requirements that must be met when dealing with an `ActorQueue`:
```
bachand, line 92: Maybe "few" -> "couple"?

README.md
```diff
@@ -89,13 +89,16 @@ FIFO execution has a key downside: the queue must wait for all previously enqueu
...
+1. The lifecycle of any `ActorQueue` should not exceed the lifecycle of its `actor`. It is strongly recommended that an `ActorQueue` be a `let` constant on the adopted `actor`. Enqueuing a task to an `ActorQueue` isntance after its adopted `actor` has been deallocated will result in a crash.
```
bachand, line 93: "isntance" -> "instance"

README.md
bachand: Since there are two.

README.md
```diff
@@ -89,13 +89,16 @@ FIFO execution has a key downside: the queue must wait for all previously enqueu
...
+2. An `actor` utilizing an `ActorQueue` should set the adopted execution context of the queue to `self` within the `actor`’s `init`. Failing to set an adopted execution context prior to enqueuing work on an `ActorQueue` will result in a crash.
 
 An `ActorQueue` can easily enqueue tasks that execute on an actor’s isolated context from a nonisolated context in order:
 ```swift
 func testActorQueueOrdering() async {
     actor Counter {
         init() {
+            // Adopting the execution context in `init` satisfies requirement #2 above.
```
bachand, line 101: Nice!

README.md
dfed: 😅 was leaving room for me to add more later, but that's certainly a silly approach. Will fix this up in #13

PR #12, opened 2023-02-12
async -> enqueue. await -> enqueueAndWait

### LINKED ISSUE #10: Rename `async` and `await` on both `FIFOQueue` and `ActorQueue`
**Goals**
* Ensure syntax highlighting does not get confused by avoiding the use of language keywords in API

**Non-Goals**
* Avoid a breaking API change

**Discussion**
Our current `async`/`await` API was inspired by both `DispatchQueue`'s `async`/`sync` APIs as well as Swift Concurrency Syntax's `async` and `await` keywords. However, Xcode's syntax highlighting gets confused by this API choice:

screenshot of Xcode syntax highlighting showing `await` as if it were a language keyword rather than a method call

As such, I'd like to change the name of these APIs.

@ktoso's [discussion of creating ordered tasks from nonisolated contexts](https://forums.swift.org/t/task-is-order-of-task-execution-deterministic/51553/31) has me thinking of two possibilities for replacing the `async` and `await` APIs in `FIFOQueue` and `ActorQueue`:

1. `enqueue` and `enqueueAndWait`
2. `send` and `sendAndWait`

I'm currently partial to the former, but am curious to get input on these options. I'm also open to alternative API suggestions!

I hope to include this name change in the `0.2.0` release of this library, which should also include the API changes in #6 and #9.

dfed:
Resolves #10

### DIFF
--- a/Sources/AsyncQueue/ActorQueue.swift
+++ b/Sources/AsyncQueue/ActorQueue.swift
@@ -35,14 +35,14 @@
 ///
 ///     nonisolated
 ///     public func log(_ message: String) {
-///         queue.async { myself in
+///         queue.enqueue { myself in
 ///             myself.logs.append(message)
 ///         }
 ///     }
 ///
 ///     nonisolated
 ///     public func retrieveLogs() async -> [String] {
-///         await queue.await { myself in myself.logs }
+///         await queue.enqueueAndWait { myself in myself.logs }
 ///     }
 ///
 ///     private let queue = ActorQueue<LogStore>()
@@ -78,7 +78,7 @@ public final class ActorQueue<ActorType: Actor> {
 
     // MARK: Public
 
-    /// Sets the actor context within which each `async` and `await`ed task will execute.
+    /// Sets the actor context within which each `enqueue` and `enqueueAndWait` task will execute.
     /// It is recommended that this method be called in the adopted actor’s `init` method.
     /// **Must be called prior to enqueuing any work on the receiver.**
     ///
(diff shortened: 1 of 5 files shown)


Gabardone (approved):
I like this naming better. The prior version had too much baggage from the other APIs that use the same names.

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -78,7 +78,7 @@ public final class ActorQueue<ActorType: Actor> {
 
     // MARK: Public
 
-    /// Sets the actor context within which each `async` and `await`ed task will execute.
+    /// Sets the actor context within which each `enqueue` and `enqueueAndWait` task will execute.
```
bachand, line 81: "`enqueue` and `enqueueAndWait`" -> "`enqueue`ed and `enqueueAndWait`ed"?

Sources/AsyncQueue/FIFOQueue.swift
```diff
@@ -54,7 +54,7 @@ public final class FIFOQueue: Sendable {
     /// Schedules an asynchronous task for execution and immediately returns.
     /// The scheduled task will not execute until all prior tasks – including suspended tasks – have completed.
     /// - Parameter task: The task to enqueue.
-    public func async(_ task: @escaping @Sendable () async -> Void) {
+    public func enqueue(_ task: @escaping @Sendable () async -> Void) {
```
bachand, line 57: Do we need to edit this `await` in this line? https://github.com/dfed/swift-async-queue/blob/3454ea6497aeb917b6fe084efd69d95d9c7cff35/Sources/AsyncQueue/FIFOQueue.swift#L25

Tests/AsyncQueueTests/FIFOQueueTests.swift
```diff
@@ -36,46 +36,46 @@ final class FIFOQueueTests: XCTestCase {
 
     // MARK: Behavior Tests
```
bachand, line 37: There are at least a couple places in the file like this that we should clean up.
https://github.com/dfed/swift-async-queue/blob/3454ea6497aeb917b6fe084efd69d95d9c7cff35/Tests/AsyncQueueTests/FIFOQueueTests.swift#L81

bachand (commented):
Nice. I agree it's good to pick unique names that don't overlap with Swift language names.

Sources/AsyncQueue/FIFOQueue.swift
dfed: oooooh solid find!

dfed:
Pushed feedback up into #13

PR #13, opened 2023-02-17
Make FIFOQueue, ActorQueue, and tests pass strict concurrency checking

dfed:
I made this PR by turning on strict concurrency checking, and then squashing the warnings/errors.


AsyncQueue.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'AsyncQueue'
-  s.version  = '0.2.0'
+  s.version  = '0.3.0'
```
dfed, line 3: Unfortunately this PR represents a (minority) breaking API change.

Package.swift
```diff
@@ -22,6 +22,11 @@ let package = Package(
             dependencies: []),
```
dfed, line 22: Ideally we'd force the vended target to conform to include `strict-concurrency=complete`, but [doing so would mark this repo as having unsafe build flags](https://stackoverflow.com/questions/74986278/cannot-use-package-as-dependency-with-concurrency-checking-enabled-in-package-sw). Once we can use `enableUpcomingFeature` (see below), this will get better.

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -50,8 +50,9 @@
 /// }
 /// ```
 ///
+/// - Warning: The `ActorQueue`'s conformance to `@unchecked Sendable` is safe if and only if `adoptExecutionContext(of:)` is called only from the adopted actor's `init` method.
 /// - Precondition: The lifecycle of an `ActorQueue` must not exceed that of the adopted actor.
-public final class ActorQueue<ActorType: Actor> {
+public final class ActorQueue<ActorType: Actor>: @unchecked Sendable {
```
dfed, line 55: If we don't make `ActorQueue` be `Sendable`, then we get the following error in consuming code whenever the `ActorQueue` instance is accessed from a `nonisolated` method:
```
Non-sendable type 'ActorQueue<ActorType>' in asynchronous access to actor-isolated property 'actorQueue' cannot cross actor boundary
```

This error makes sense: we're accessing a non-sendable `let` constant outside of the isolated context. I'm looking forward to Swift 6 when I don't need to manually turn on this checking to see these kinds of issues in my packages.

dfed: We also can't make this type `Sendable` because `weakExecutionContext` is mutable and therefore the type _is not Sendable_. But given the API contract of:
1. `adoptExecutionContext(of:)` must be called from `init`
2. `adoptExecutionContext(of:)` must not be called more than once
3. An `ActorQueue` should be a `private let` constant on the adopted actor instance

The `ActorQueue` is effectively Sendable, and therefore we can mark it as such.

Sources/AsyncQueue/FIFOQueue.swift
```diff
@@ -85,7 +85,7 @@ public final class FIFOQueue: Sendable {
     ///   - isolatedActor: The actor within which the task is isolated.
     ///   - task: The task to enqueue.
     /// - Returns: The value returned from the enqueued task.
-    public func enqueueAndWait<ActorType: Actor, T>(on isolatedActor: isolated ActorType, _ task: @escaping @Sendable (isolated ActorType) async -> T) async -> T {
+    public func enqueueAndWait<ActorType: Actor, T: Sendable>(on isolatedActor: isolated ActorType, _ task: @escaping @Sendable (isolated ActorType) async -> T) async -> T {
```
dfed, line 88: This `T: Sendable` conformance was only required on methods that execute on an isolated actor. The warning we saw by not having this conformance was:
```
Non-sendable type 'T' returned by implicitly asynchronous call to actor-isolated function cannot cross actor boundary
```

This warning makes sense: `T` is not `Sendable`, and therefore how could we be passing it across contexts. However, the fact that this warning only surfaced when we explicitly utilizing an isolated parameter seems like a bug – in all cases we are sending `T` from an actor's isolated context to the current context. So I made all return types `T` conform to `Sendable`.

AsyncQueue.podspec
dfed, line 3: Unfortunately this PR represents a (reasonably small) breaking API change.

Tests/AsyncQueueTests/ActorQueueTests.swift
```diff
@@ -65,17 +65,17 @@ final class ActorQueueTests: XCTestCase {
 
     func test_enqueue_taskParameterIsAdoptedActor() async {
         let semaphore = Semaphore()
-        systemUnderTest.enqueue { counter in
-            XCTAssertTrue(counter === self.counter)
```
dfed, line 69: capturing non-sendable `self` in the sendable closure is not good! So instead we now capture the `counter` directly.

Package.swift
bachand: OK I see that we use unsafe build flags in the test target and that's OK since it's not being consumed by other packages.

Sources/AsyncQueue/FIFOQueue.swift
bachand: To make sure I follow, we got a warning in all cases except the ones where the `task` closure doesn't have an `isolated` parameter? Making the return type `Sendable` seems reasonable but I want to try to understand better which methods didn't have a warning so I can try to understand why some methods are warning differently than others.

bachand: If I'm following correctly I think that these two methods on `FIFOQueue` didn't have warnings.

```swift
public func enqueueAndWait<T: Sendable>(_ task: @escaping @Sendable () async -> T) async -> T
public func enqueueAndWait<T: Sendable>(_ task: @escaping @Sendable () async throws -> T) async throws -> T
```

Is that right?

bachand (approved):
One question but overall LGTM. Thanks for addressing feedback from previous PRs.

Sources/AsyncQueue/FIFOQueue.swift
dfed: > To make sure I follow, we got a warning in all cases except the ones where the task closure doesn't have an `isolated` parameter?

Correct!

> Making the return type `Sendable` seems reasonable but I want to try to understand better which methods didn't have a warning so I can try to understand why some methods are warning differently than others.

My best guess is that this is a compiler bug. In all the methods I marked we're crossing a task boundary: from the task that created the async for loop to the call-site. But only in the methods where we have an `isolated` parameter does the compiler notice that we're crossing a task boundary and therefore only then does it require `Sendable` conformance.

> Is that right?

That is absolutely right.

PR #14, opened 2023-02-25
Create a MainActorQueue

dfed:
Introduces an `ActorQueue`-like `MainActorQueue` that targets the `@MainActor`. While theoretically this could have been implemented by utilizing an `ActorQueue` under the hood and adopting the execution context of `MainActor.shared`, it seemed more straightforward to create a new implementation where even the async for-loop ran on the main actor.

It might be useful to also include a `MainFIFOQueue` in this package, but I'll save that for another day.


Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -154,7 +154,7 @@ public final class ActorQueue<ActorType: Actor>: @unchecked Sendable {
 }
 
 extension Actor {
-    func suspendUntilStarted(_ task: @escaping @Sendable (isolated Self) async -> Void) async {
+    fileprivate func suspendUntilStarted(_ task: @escaping @Sendable (isolated Self) async -> Void) async {
```
dfed, line 157: not a required change at all but since I don't have an expectation this method to be used outside of this file I might as well lock it down.

Sources/AsyncQueue/MainActorQueue.swift
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    deinit {
+        taskStreamContinuation.finish()
+    }
+
+    // MARK: Public
+
+    /// An easily accessible instance of a `MainActorQueue`.
+    public static let shared = MainActorQueue()
```
dfed, line 56: I'm tempted to make a free function `MainTask` that takes a `Sendable @MainActor () async -> Void` as an argument and executes the closure on `MainActorQueue.shared`. Basically a `Task { @MainActor` except that it runs on the `MainActorQueue` so you can guarantee ordering.

README.md
```diff
@@ -129,6 +129,46 @@ func testActorQueueOrdering() async {
 }
 ```
 
+### Sending ordered asynchronous tasks to the `@MainActor` from a nonisolated context
+
+Use a `MainActorQueue` to send ordered asynchronous tasks to the `@MainActor`’s isolated context from nonisolated or synchronous contexts. Tasks sent to this queue type are guaranteed to begin executing in the order in which they are enqueued. Like an `ActorQueue`, execution order is guaranteed only until the first [suspension point](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html#ID639) within the enqueued task. A `MainActorQueue` executes tasks within the its adopted actor’s isolated context, resulting in `MainActorQueue` task execution having the same properties as a `@MainActor`'s' code execution: code between suspension points is executed atomically, and tasks sent to a single `MainActorQueue` can await results from the queue without deadlocking.
```
Gabardone, line 134:
```suggestion
Use a `MainActorQueue` to send ordered asynchronous tasks to the `@MainActor`’s isolated context from nonisolated or synchronous contexts. Tasks sent to this queue type are guaranteed to begin executing in the order in which they are enqueued. Like an `ActorQueue`, execution order is guaranteed only until the first [suspension point](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html#ID639) within the enqueued task. A `MainActorQueue` executes tasks within its adopted actor’s isolated context, resulting in `MainActorQueue` task execution having the same properties as a `@MainActor`'s' code execution: code between suspension points is executed atomically, and tasks sent to a single `MainActorQueue` can await results from the queue without deadlocking.
```
typo.

Gabardone: Also not sure if the second half of that explanation is needed, may work out better to just lead with the behavior rather than re-explaining `ActorQueue`

Sources/AsyncQueue/MainActorQueue.swift
Gabardone: How often would you want to manage multiple `MainActorQueue` instances?

Sources/AsyncQueue/MainActorQueue.swift
```diff
@@ -0,0 +1,110 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+/// A queue that enables enqueing ordered asynchronous tasks from a nonisolated context onto the `@MainActor`'s isolated context.
+/// Tasks are guaranteed to begin executing in the order in which they are enqueued. However, if a task suspends it will allow subsequently enqueued tasks to begin executing.
+/// This queue exhibits the execution behavior of an actor: tasks sent to this queue can re-enter the queue, and tasks may execute in non-FIFO order when a task suspends.
+///
+/// A `MainActorQueue` ensures tasks sent from a nonisolated context to the `@MainActor`'s isolated context begin execution in order.
+public final class MainActorQueue: Sendable {
```
Gabardone, line 28: We're not at the point where we could just have `MainActorQueue` be a special case of the regular actor queue could we?

Gabardone (approved):
I'm not seeing anything suspicious. A bit bummed that we can't have this be just a special case of `ActorQueue` but I'm sure you've already looked into that.

Sources/AsyncQueue/MainActorQueue.swift
dfed: It *could* be a special case of `ActorQueue`, but the biggest difference is that the `MainActorQueue` runs its `for await task in taskStream` _on `Main`_ rather than on some background queue. Limiting queue hops makes us slightly more efficient / more akin to `DispatchQueue.main.async` in terms of performance.

We also mark every closure as being a `@MainActor` closure, which helps ensure the compiler does the right thing / always keeps us on `Main`. (I recently found an issue where lack of explicitness cause some code to run off of `Main` that should have been running on `Main`, so I default now to "be explicit if possible")

dfed: > Limiting queue hops makes us slightly more efficient / more akin to `DispatchQueue.main.async` in terms of performance.

I just ran some tests and `MainActorQueue` still manages to have a queue hop from the `Task { @MainActor in` call – the `Task` initially starts running on a background queue and then enters `Main` before executing our code (per the debugger), so we're still slower than `DispatchQueue.main.async`. Maaaaybe I take a pass at making this implemented with an `ActorQueue` 🤔

Sources/AsyncQueue/MainActorQueue.swift
dfed: I don't think you'd really ever want to manage multiple instances, since ordering between instances is not guaranteed which could lead to confusion.

README.md
dfed: Good catch on the typo! And fair that the documentation is repetitive here but sometimes repetition is helpful: I'm trying to drive home that *this type behaves the same way an `ActorQueue` does*, and also trying to help folk who might only be using a `MainActorQueue` and not utilizing the other types – you shouldn't have to read the whole README to have an understanding of what's happening in this type imo.

Sources/AsyncQueue/MainActorQueue.swift
dfed: I also just tried to implement `MainActorQueue` utilizing an `ActorQueue` and remembered why this doesn't work:
`ActorQueue` adopts the execution context of an `Actor`-conforming type, and `final class MainActorQueue` does not conform to `Actor`!

Heck, even `@MainActor final class MainActorQueue` yields a `Non-actor type 'MainActorQueue' cannot conform to the 'Actor' protocol` error.

Sources/AsyncQueue/MainActorQueue.swift
dfed: Given the above I just made the `init` for `MainActorQueue` internal: 1d4842d

Creating a `MainTask` type is out of scope – I just spiked on it and it's currently more effort than it's worth.

Sources/AsyncQueue/MainActorQueue.swift
Gabardone: It's ok, this might be doable a few Swift revisions from now but it's clear we're going to be hitting some pain points.

README.md
Gabardone: I get what you mean. As a possible alternative, is there some way to link to the original in the same README?

Not sure how this particular breed of markdown works here 😬

dfed: Yes, but only the header which is a bit far from the thing I want to describe

Sources/AsyncQueue/MainActorQueue.swift
dfed: So, actually doable:
```
diff --git a/Sources/AsyncQueue/MainActorQueue.swift b/Sources/AsyncQueue/MainActorQueue.swift
index 3dde0b7..5a4adf0 100644
--- a/Sources/AsyncQueue/MainActorQueue.swift
+++ b/Sources/AsyncQueue/MainActorQueue.swift
@@ -31,23 +31,7 @@ public final class MainActorQueue: Sendable {
 
     /// Instantiates a main actor queue.
     init() {
-        var capturedTaskStreamContinuation: AsyncStream<@Sendable @MainActor () async -> Void>.Continuation? = nil
-        let taskStream = AsyncStream<@Sendable @MainActor () async -> Void> { continuation in
-            capturedTaskStreamContinuation = continuation
-        }
-        // Continuation will be captured during stream creation, so it is safe to force unwrap here.
-        // If this force-unwrap fails, something is fundamentally broken in the Swift runtime.
-        taskStreamContinuation = capturedTaskStreamContinuation!
-
-        Task.detached { @MainActor in
-            for await task in taskStream {
-                await MainActor.shared.suspendUntilStarted(task)
-            }
-        }
-    }
-
-    deinit {
-        taskStreamContinuation.finish()
+        actorQueue.adoptExecutionContext(of: MainActor.shared)
     }
 
     // MARK: Public
@@ -59,7 +43,9 @@ public final class MainActorQueue: Sendable {
     /// The scheduled task will not execute until all prior tasks have completed or suspended.
     public func enqueue(_ task: @escaping @Sendable @MainActor () async -> Void) {
-        taskStreamContinuation.yield(task)
+        actorQueue.enqueue { _ in
+            await task()
+        }
     }
 
     /// Schedules an asynchronous task and returns after the task is complete.
@@ -67,10 +53,8 @@ public final class MainActorQueue: Sendable {
     public func enqueueAndWait<T: Sendable>(_ task: @escaping @Sendable @MainActor () async -> T) async -> T {
-            taskStreamContinuation.yield {
-                continuation.resume(returning: await task())
-            }
+        await actorQueue.enqueueAndWait { _ in
+            await task()
         }
     }
 
@@ -79,33 +63,12 @@ public final class MainActorQueue: Sendable {
     public func enqueueAndWait<T: Sendable>(_ task: @escaping @Sendable @MainActor () async throws -> T) async throws -> T {
-        try await withUnsafeThrowingContinuation { continuation in
-            taskStreamContinuation.yield {
-                do {
-                    continuation.resume(returning: try await task())
-                } catch {
-                    continuation.resume(throwing: error)
-                }
-            }
+        try await actorQueue.enqueueAndWait { _ in
+            try await task()
         }
     }
 
     // MARK: Private
 
-    private let taskStreamContinuation: AsyncStream<@Sendable @MainActor () async -> Void>.Continuation
-}
-
-extension MainActor {
-    @MainActor
-    fileprivate func suspendUntilStarted(_ task: @escaping @Sendable @MainActor () async -> Void) async {
-        // Suspend the calling code until our enqueued task starts.
-            // Utilize the serial (but not FIFO) @MainActor context to execute the task without requiring the calling method to wait for the task to complete.
-            Task { @MainActor in
-                // Signal that the task has started. Since this `task` is executing on the main actor's execution context, the order of execution is guaranteed.
-                continuation.resume()
-                await task()
-            }
-        }
-    }
+    private let actorQueue = ActorQueue<MainActor>()
 }
```
But def more queue hops – will keep this as-is for now.

PR #16, opened 2023-04-09
Bump Xcode

dfed:
This PR:
1. Bumps our Swift version to 5.8 in the Package.swift to enable adopting `enableUpcomingFeature`
2. Adopt `fulfillment(of:timeout:)` as `waitForExpectations(timeout:)` is deprecated in an async context
3. Bumps Xcode version to 14.3.0
4. Update macOS version to v13
5. Updates our ruby version to work with macOS 13 CI

This PR will fail CI until https://github.com/ruby/setup-ruby/issues/500 is resolved


.github/workflows/ci.yml
```diff
@@ -9,22 +9,19 @@ on:
 jobs:
   pod-lint:
     name: Pod Lint
-    runs-on: macOS-12
+    runs-on: macos-13
     steps:
-      - uses: ruby/setup-ruby@v1
-        with:
-          ruby-version: '2.7.6'
```
dfed, line 16: we no longer need these lines because we're using a built-in ruby version rather than a custom one.

Gabardone (approved):
LGTM 👍🏽

PR #19, opened 2023-09-25
Resolve Xcode 15 warnings

dfed:
Simple stuff. Xcode 15 caught a non-Sendable boundary crossing issue that Xcode 14 did not in tests.

### DIFF
--- a/Tests/AsyncQueueTests/SemaphoreTests.swift
+++ b/Tests/AsyncQueueTests/SemaphoreTests.swift
@@ -118,7 +118,7 @@ private extension Semaphore {
     /// Enqueues an asynchronous task and increments a counter after the task completes.
     /// This method suspends the caller until the asynchronous task has begun, ensuring ordered execution of enqueued tasks.
     /// - Parameter task: A unit of work that returns work to execute after the task completes and the count is incremented.
-    func enqueueAndCount(using counter: UnsafeCounter, _ task: @escaping @Sendable (isolated Semaphore) async -> ((isolated Semaphore) -> Void)?) async {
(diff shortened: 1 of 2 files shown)


Gabardone (approved):
👍🏽

Tests/AsyncQueueTests/SemaphoreTests.swift
```diff
@@ -118,7 +118,7 @@ private extension Semaphore {
     /// Enqueues an asynchronous task and increments a counter after the task completes.
     /// This method suspends the caller until the asynchronous task has begun, ensuring ordered execution of enqueued tasks.
     /// - Parameter task: A unit of work that returns work to execute after the task completes and the count is incremented.
-    func enqueueAndCount(using counter: UnsafeCounter, _ task: @escaping @Sendable (isolated Semaphore) async -> ((isolated Semaphore) -> Void)?) async {
+    func enqueueAndCount(using counter: UnsafeCounter, _ task: @escaping @Sendable (isolated Semaphore) async -> (@Sendable (isolated Semaphore) -> Void)?) async {
```
bachand, line 121: If I'm understanding correctly, we needed to make this change since this method is defined on `Semaphore` which is an actor. Is that the right way to think about it?

bachand (commented):
LGTM!

Tests/AsyncQueueTests/SemaphoreTests.swift
dfed: The closure being returned is crossing a thread boundary when we `await` the result from an actor, so it needs to be Sendable.

PR #21, opened 2023-10-13
Minor cleanup in README and package file

dfed:
This MR does two things:
1. Adds `StrictConcurrency` checking to the main project
2. Removes unnecessary `nonisolated` decorators from `async` methods – if the method is `async` it doesn't matter to the caller if it's isolated or not.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -19,7 +19,10 @@ let package = Package(
     targets: [
         .target(
             name: "AsyncQueue",
-            dependencies: []),
+            dependencies: [],
+            swiftSettings: [
+                .enableUpcomingFeature("StrictConcurrency")
+            ]),
         .testTarget(
             name: "AsyncQueueTests",
             dependencies: ["AsyncQueue"],
--- a/README.md
+++ b/README.md
@@ -60,7 +60,6 @@ func testFIFOQueueOrdering() async {
             }
         }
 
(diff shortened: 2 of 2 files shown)


Gabardone (approved):
Well if there's one project that can benefit from strict concurrency checking it's this one.

Package.swift
```diff
@@ -19,7 +19,10 @@ let package = Package(
     targets: [
         .target(
             name: "AsyncQueue",
-            dependencies: []),
+            dependencies: [],
+            swiftSettings: [
+                .enableUpcomingFeature("StrictConcurrency")
```
bachand, line 24: 👍

README.md
```diff
@@ -60,7 +60,6 @@ func testFIFOQueueOrdering() async {
             }
         }
 
-        nonisolated
```
bachand, line 63: @dfed should we update this too?

https://github.com/dfed/swift-async-queue/blob/6837ecbeb394eb1fc7ec6b6a858636074309abd9/Sources/AsyncQueue/ActorQueue.swift#L43-L46

bachand (commented):
Nice!

README.md
dfed: Yes we should! Good catch. I'll push this change up directly.

PR #25, opened 2024-01-11
Add support for more Swift and Xcode versions

dfed:


dfed:
Since we'd need to branch our code paths to support older Swift versions, I'm going to close this out since I haven't seen demand for this. I was mostly adding this to see if I could get more green checkmarks on swift package index 😅, which isn't really the best of goals.

PR #26, opened 2024-01-31
Run CI against visionOS

dfed:
As title!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,25 +9,26 @@ on:
 jobs:
   pod-lint:
     name: Pod Lint
-    runs-on: macos-13
+    runs-on: macos-14
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_15.0.1.app/Contents/Developer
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -9,19 +9,19 @@ on:
 jobs:
   pod-lint:
     name: Pod Lint
-    runs-on: macos-13
+    runs-on: macos-14
```
dfed, line 12: migrating to macOS 14 [which is now ARM](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/) while we're here

.github/workflows/ci.yml
```diff
@@ -9,19 +9,19 @@ on:
...
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
       - name: Bundle Install
         run: bundle install
       - name: Select Xcode Version
-        run: sudo xcode-select --switch /Applications/Xcode_15.0.1.app/Contents/Developer
+        run: sudo xcode-select --switch /Applications/Xcode_15.2.app/Contents/Developer
```
dfed, line 19: Xcode 15.0.1 on macOS 14 is pretty buggy and has a lot of spurious failures. Going to Xcode 15.2 to try to avoid this, plus we need 15.2 for visionOS.

PR #27, opened 2024-01-31
Add visionOS support to podspec

dfed:


Gemfile
```diff
@@ -2,4 +2,4 @@ ruby '3.0.6'
 
 source 'https://rubygems.org'
 
-gem 'cocoapods', '~> 1.14.0'
+gem 'cocoapods', '~> 1.15.0'
```
dfed, line 5: We need latest cocoa pods in order to publish a spec that includes visionOS

PR #39, opened 2024-09-23
Adopt Swift 6. Drop support for Xcode 15 and below

dfed:
This is a breaking change.

In addition to updating our tooling, I have:
1. Simplified `MainActorQueue` and `ActorQueue` now that Swift 6 enabled ordered task enqueueing from a known concurrency context
2. Updated the README to explain a little better how this library is still useful given that Swift 6 has improved _some_ of the nondeterminism when hopping onto an actor queue.


README.md
```diff
@@ -9,11 +9,10 @@ A library of queues that enable sending ordered tasks from synchronous to asynch
 
 ## Task Ordering and Swift Concurrency
 
-Tasks sent from a synchronous context to an asynchronous context in Swift Concurrency prior to Swift 6 are inherently unordered. Consider the following test:
+Tasks sent from a synchronous context to an asynchronous context in Swift Concurrency are inherently unordered. Consider the following test:
 
 ```swift
-@MainActor
-func testMainActorTaskOrdering() async {
+func testActorTaskOrdering() async {
```
dfed, line 15: this is tricky, but all we need to break this test in Swift 6 is remove the global actor.

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -147,16 +153,9 @@ public final class ActorQueue<ActorType: Actor>: @unchecked Sendable {
...
-        await withUnsafeContinuation { continuation in
-            // Utilize the serial (but not FIFO) Actor context to execute the task without requiring the calling method to wait for the task to complete.
-            Task {
-                // Signal that the task has started. Since this `task` is executing on the current actor's execution context, the order of execution is guaranteed.
-                continuation.resume()
-                await task(self)
-            }
-        }
-    }
+/// A global actor used for synchronizing task execution.
```
dfed, line 156: I figure if we can avoid clogging the main queue, we should.

PR #40, opened 2024-10-07
Remove need for ActorQueueSynchronization

dfed:
We don't need to bounce through a global queue in order to ensure FIFS (first-enqueued-first-start) behavior in our ActorQueue. Instead, we just can make the async for loop await switching to the ActorType's context.

This PR is unlikely to affect performance. This PR should make this code a bit easier to follow, however.

### DIFF
--- a/Sources/AsyncQueue/ActorQueue.swift
+++ b/Sources/AsyncQueue/ActorQueue.swift
@@ -78,7 +78,7 @@ public final class ActorQueue<ActorType: Actor>: @unchecked Sendable {
 
         actor.execute { [taskStream] _ in
             func beginExecuting(
-                _ operation: sending @escaping (isolated ActorType) async -> Void,
+                _ operation: @escaping (isolated ActorType) async -> Void,
                 in context: isolated ActorType
             ) {
                 // In Swift 6, a `Task` enqueued from an actor begins executing immediately on that actor.
--- a/Sources/AsyncQueue/ActorQueue.swift
+++ b/Sources/AsyncQueue/ActorQueue.swift
@@ -89,6 +89,8 @@ public final class ActorQueue<ActorType: Actor>: @unchecked Sendable {
             }
 
             for await actorTask in taskStream {
+                // The compiler does not have a guarantee that `actor === actorTask.executionContext`, so we must `await`.
+                // In practice, however, that condition is always true, and therefore we do not end up suspending here.
                 await beginExecuting(
                     actorTask.task,
                     in: actorTask.executionContext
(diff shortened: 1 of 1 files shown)


Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -84,10 +71,30 @@ public final class ActorQueue<ActorType: Actor>: @unchecked Sendable {
     /// **Must be called prior to enqueuing any work on the receiver.**
     ///
     /// - Parameter actor: The actor on which the queue's task will execute. This parameter is not retained by the receiver.
-    /// - Warning: Calling this method more than once will result in an assertion failure.
+    /// - Precondition: Calling this method more than once will result in a precondition failure.
```
dfed, line 74: Now that we're kicking off the stream execution here, `precondition` feels appropriate.

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -84,10 +71,30 @@ public final class ActorQueue<ActorType: Actor>: @unchecked Sendable {
...
+                }
+            }
+
+            for await actorTask in taskStream {
+                await beginExecuting(
+                    actorTask.task,
+                    in: actorTask.executionContext
+                )
+            }
+        }
```
dfed, line 97: this entire closure executes within the `actor` context!

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -153,9 +161,13 @@ public final class ActorQueue<ActorType: Actor>: @unchecked Sendable {
...
 
-/// A global actor used for synchronizing task execution.
-@globalActor
-private struct ActorQueueSynchronization {
-    fileprivate actor Synchronization {}
-    fileprivate static let shared = Synchronization()
+extension Actor {
+    nonisolated
+    func execute(
+        _ task: @escaping @Sendable (isolated Self?) async -> Void
```
dfed, line 167: a bit tacky but enables us to isolate the closure to the actor's context without retaining the actor.

Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -84,10 +71,30 @@ public final class ActorQueue<ActorType: Actor>: @unchecked Sendable {
...
+            ) {
+                // In Swift 6, a `Task` enqueued from an actor begins executing immediately on that actor.
+                // Since we're running on our actor's context due to the isolated parmater, we can just dispatch a Task to get first-enqueued-first-start task execution.
+                Task {
+                    await operation(context)
+                }
+            }
+
+            for await actorTask in taskStream {
+                await beginExecuting(
```
dfed, line 92: we need to `await` here despite the fact that we're already isolated to the `context` because the compiler does not have a guarantee that `actorTask.executionContext === `actor`. In practice, these are always the same and therefore we do not suspend here.

PR #41, opened 2024-10-20
Adopt Swift Testing

dfed:
Migrates from XCTest to [Swift Testing](https://developer.apple.com/documentation/testing). I ended up having to create an `Expectation` type to mimic what XCTest has, since [confirmations](https://developer.apple.com/documentation/testing/testing-asynchronous-code) require the `confirmation` to be confirmed before the confirmation closure exits, which doesn't work terribly well in a highly async environment.

### DIFF
--- a/Tests/AsyncQueueTests/Utilities/Expectation.swift
+++ b/Tests/AsyncQueueTests/Utilities/Expectation.swift
@@ -22,7 +22,6 @@
 
 import Testing
 
-@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
 public actor Expectation {
 
     // MARK: Initialization
@@ -46,6 +45,7 @@ public actor Expectation {
 
     // MARK: Public
 
+    @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
     public func fulfillment(
         within duration: Duration,
         filePath: String = #filePath,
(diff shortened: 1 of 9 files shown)


Tests/AsyncQueueTests/Utilities/Expectation.swift
```diff
@@ -0,0 +1,120 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Testing
+
+public actor Expectation {
```
dfed, line 25: This is quite similar to the `Expectation` in the [dfed/swift-testing-expectation](https://github.com/dfed/swift-testing-expectation) repo. I figure it is best to keep this repo having zero dependencies, and a little copy/paste never hurt anyone.

Issue #50, opened 2025-03-17
Cancellable interface for FIFOQueue


vladyslavsosiuk:
**Goals**
- Introduce a cancellable interface to the existing FIFOQueue implementation.

**Technical Details**
- Tasks can be cancelled before they start, and the cancellation will be propagated correctly.
- Task cancellation is handled using `withTaskCancellationHandler` to ensure proper cleanup.

**API Overview**
```swift
    public func enqueue<T: Sendable>(
        _ task: @escaping @Sendable () async throws -> T
    ) async throws -> T
```

**Example Usage**
```swift
let queue = FIFOQueue()

let task = Task {
    do {
        let result = try await queue.enqueueWithCancellation {
            try Task.sleep(seconds: 1)
        }
        print("Task completed with result: \(result)")
    } catch {
        print("Task was cancelled or failed with error: \(error)")
    }
}
task.cancel()
```

I have developed a similar queue functionality that has a cancellable `enqueue` method implementation. This allowed tasks to be enqueued with the ability to cancel them if needed either when they already started or when they are waiting in the queue. I believe this feature could be a valuable addition to your library. Would you be interested in me opening a pull request with this functionality?

**Update:**
I discovered that while implementing cancellation is possible, it loses the ability to propagate Task locals to the enqueued tasks. Extra work is required to hold a handle on a task to cancel it without starting it immediately and this involves starting a task from the escaped boundary, which means it doesn't retain locals of the tasks that enqueued the task. I believe this behavior would be unexpected and confusing with the proposed interface moving forward.

dfed:
I like the idea of adding cancellation! I've played with it locally over the years but haven't prioritized because I haven't hit this need myself.

This repo's queues are very much in the world of [Unstructured Concurrency](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/concurrency/#Unstructured-Concurrency). I don't think we necessarily want to introduce structured concurrency where tasks created within a particular context can all be cancelled in one fell swoop.

Thoughts on an API that looks like:
```swift
public func enqueue<Success: Sendable, Failure: Error>(_ task: @escaping @Sendable () async -> Void) -> FIFOTask<Success, Failure>
```

Where the API for `FIFOTask` looks vaguely like a `Task`:

```swift
public struct FIFOTask<Success, Failure> {
    public var value: Success { get async throws }

    public var isCancelled: Bool { get }

    public func cancel()
}

extension FIFOTask where Failure == Never {
[...]

PR #51, opened 2025-04-02
Create Task API -> 0.7.0

dfed:
Addresses #50


Sources/AsyncQueue/ActorQueue.swift
```diff
@@ -101,65 +102,246 @@ public final class ActorQueue<ActorType: Actor>: @unchecked Sendable {
...
+    /// it only makes it impossible for you to explicitly cancel the task.
+    ///
+    /// - Parameters:
+    ///   - priority: The priority of the task.
+    ///     Pass `nil` to use the priority from `Task.currentPriority`.
+    ///   - actorQueue: The queue on which to enqueue the task.
+    ///   - operation: The operation to perform.
+    @discardableResult
+    public init<ActorType: Actor>(
+        priority: TaskPriority? = nil,
```
dfed, line 169: should remove... priority is the same as queue priority

dfed: now we're respecting it! removed from `FIFOQueue` though where it makes less sense

PR #59, opened 2025-09-01
Use NSLock for better performance

dfed:


dfed:
eh. honestly not sure this is worth it given linux+android support.

PR #61, opened 2025-11-27
queue task add name

songhuaixu:


Sources/AsyncQueue/FIFOQueue.swift
```diff
@@ -27,12 +27,13 @@ public final class FIFOQueue: Sendable {
 	// MARK: Initialization
 
 	/// Instantiates a FIFO queue.
+	/// - Parameter name: Human readable name of the task.
```
dfed:
```suggestion
	/// - Parameter name: Human readable name of the queue.
```

dfed (commented):
Seems like a reasonable addition once it's compiling.

I think it'd make sense to add this `name` parameter to the `Task` initializers below as well, so that each individual task (not just the top-level queue) can be named.

dfed:
Good improvements. Not sure why CI is still failing. Might be Swift 6.0 vs 6.2? I'm on mobile today so not sure. Can look closer tomorrow if you're stuck. Let me know!

songhuaixu:
But [Task init name](https://developer.apple.com/documentation/swift/task/init(name:priority:operation:)-2dll5)  has long supported it

Issue #62, opened 2025-11-28
Proposal: Add cancelAll() method to FIFOQueue/ActorQueue


songhuaixu:
**Goals** 

Provide a thread-safe method to cancel all pending operations in a queue

**Non-Goals**
No changes to Queue's fundamental execution model

Does not handle operation dependencies automatically

**Investigations**

Individual cancellation: Requires maintaining references to all Task and calling cancel() on each, which is cumbersome

songhuaixu:
It's not quite what I imagined, but it's still amazing.

PR #63, opened 2025-11-28
Create CancellableQueue

dfed:
Addresses #62 by creating a new `CancellableQueue` type that wraps either `FIFOQueue` or `ActorQueue` and allows tasks enqueued on the cancellable queue to be cancelled.

I created a separate queue type to avoid making every `FIFOQueue` and `ActorQueue` create yet another queue – this feels like a specialized use case, and so a specialized queue type seemed like a reasonable solution.

### DIFF
--- a/Sources/AsyncQueue/CancellableQueue.swift
+++ b/Sources/AsyncQueue/CancellableQueue.swift
@@ -25,8 +25,6 @@ import Foundation
 /// A queue wrapper that enables cancelling all currently executing and pending tasks.
 ///
 /// A `CancellableQueue` wraps either a `FIFOQueue` or an `ActorQueue` and tracks all tasks enqueued on it.
-/// Calling `cancelTasks()` will cancel both the currently executing task and any tasks waiting in the queue.
-/// Tasks that have already completed are unaffected, and tasks enqueued after `cancelTasks()` is called will execute normally.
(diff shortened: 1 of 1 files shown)


Sources/AsyncQueue/CancellableQueue.swift
```diff
@@ -0,0 +1,448 @@
...
+	/// Instantiates a cancellable queue.
+	/// - Parameter underlyingQueue: The queue whose work can be cancelled.
+	public init(underlyingQueue: UnderlyingQueue) {
+		self.underlyingQueue = underlyingQueue
+	}
+
+	// MARK: Public
+
+	/// Cancels the currently executing task, as well as any task currently pending in the queue.
+	public func cancelTasks() {
```
dfed: `cancelAll` is a reasonable name too. I like `cancelTasks` a little bit better because it doesn't imply that future tasks will be cancelled as well.

Sources/AsyncQueue/CancellableQueue.swift
```diff
@@ -0,0 +1,448 @@
...
+	init(value: State) {
+		unsafeValue = value
+	}
+
+	// MARK: Fileprivate
+
+	fileprivate var value: State {
+		lockQueue.sync { unsafeValue }
+	}
+
```
dfed:
```suggestion
```
Unused!

PR #76, opened 2026-06-09
Use FoundationEssentials in CancellableQueue when available

lhoward:
## Motivation

`CancellableQueue.swift` is the only file in `AsyncQueue` that does a plain `import Foundation`. On Linux that single import pulls `libFoundation.so` → `libFoundationInternationalization.so` → `_FoundationICU` into the linked binary via the autolink directive — even though the file only uses `UUID` and `DispatchQueue`, and no ICU/internationalization functionality is ever called.

For size-constrained / embedded Linux targets this drags in the entire ICU data blob (tens of MB of flash) and the Foundation internationalization library for no functional benefit.

## Change

```swift
#if canImport(FoundationEssentials)
import FoundationEssentials   // UUID
import Dispatch               // DispatchQueue
#else
import Foundation
#endif
```

- `UUID` is provided by `FoundationEssentials`.
- `DispatchQueue` comes from `Dispatch` (already imported directly in `Utilities/Delivery.swift`, so this matches existing style in the package).
- Platforms without `FoundationEssentials` fall back to full `Foundation` exactly as before — no behavior change.

No public API change; `CancellableQueue` behaves identically. This just lets it link without full Foundation/ICU where `FoundationEssentials` exists.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/AsyncQueue/CancellableQueue.swift
+++ b/Sources/AsyncQueue/CancellableQueue.swift
@@ -20,7 +20,14 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-import Foundation
+#if canImport(FoundationEssentials)
+	import FoundationEssentials
+#else
+	import Foundation
+#endif
+#if canImport(Dispatch)
+	import Dispatch
+#endif
 
 /// A queue wrapper that enables cancelling all currently executing and pending tasks.
 ///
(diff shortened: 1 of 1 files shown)


Sources/AsyncQueue/CancellableQueue.swift
```diff
@@ -20,7 +20,12 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+#if canImport(FoundationEssentials)
```
dfed, line 23: For good measure can we also add a `canImport(Dispatch)` here? I know that this is tautological today but I’d like to be defensive against future potential breakages.

dfed (approved):
Thank you for the contribution! One nit before I merge assuming CI passes.

Sources/AsyncQueue/CancellableQueue.swift
lhoward: OK, I'll do this as follows:

```swift
#if canImport(FoundationEssentials)
import FoundationEssentials
#else
import Foundation
#endif
#if canImport(Dispatch)
import Dispatch
#endif
```

think it looks a bit neater than nested import

dfed: That works too!

